PR #46, opened 2022-12-15
Vctr 63 implement sum min max maxabs

### LINKED ISSUE #11: Implement sum, min, max, maxAbs

JanosGit:


include/vctr/Expressions/Core/Max.h
```diff
@@ -0,0 +1,184 @@
...
+template <size_t extent, class SrcType>
+requires std::totally_ordered<ValueType<SrcType>>
+class Max : public ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (Max)
+
+    static constexpr value_type reductionResultInitValue = std::numeric_limits<value_type>::min();
```
TheWaveWarden: For floating types, `min()` is not what we want here, but rather `std::numeric_limits<type>::lowest()` 🙂

include/vctr/Expressions/Core/Max.h
```diff
@@ -0,0 +1,184 @@
...
+template <size_t extent, class SrcType>
+requires std::totally_ordered<ValueType<SrcType>>
+class MaxAbs : public ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (MaxAbs)
+
+    static constexpr value_type reductionResultInitValue = std::numeric_limits<value_type>::min();
```
TheWaveWarden: Same here!

include/vctr/Containers/VctrBase.h
```diff
@@ -366,6 +366,24 @@ class VctrBase : private Config,
     /** Subtracts a constant from this in place */
     void operator-= (value_type c);
 
+    //==============================================================================
+    // Math reduction operations
+    //==============================================================================
+    /** Returns the minimal value of all elements */
+    ElementType min() const requires std::totally_ordered<ElementType>;
```
pmglr, line 390: Interesting. How do I make sure that a home-made ElementType is totally_ordered, i.e., [that the comparison operators ==,!=,<,>,<=,>= on a type yield results consistent with a strict total order on the type](https://en.cppreference.com/w/cpp/concepts/totally_ordered)?

pmglr, line 390: Please terminate doxygen comments with a period for good style. :)

include/vctr/Containers/VctrBase.h
```diff
@@ -366,6 +366,24 @@ class VctrBase : private Config,
...
+    /** Returns the maximum value of all elements */
+    ElementType max() const requires std::totally_ordered<ElementType>;
+
+    /** Returns the maximum absolute value of all elements */
+    ElementType maxAbs() const requires std::totally_ordered<ElementType>;
+
+    /** Returns the sum of all elements */
+    ElementType sum() const requires has::operatorPlusEquals<ElementType>;
+
 protected:
```
pmglr, line 404: This header uses //============== separators in some places. For consistency and because this is a very long header, I suggest also putting them after access specifiers although this line is not part of the PR!

include/vctr/Containers/VctrBase.tpp
```diff
@@ -107,5 +107,39 @@ void VctrBase<ElementType, StorageType, extent, StorageInfoType>::operator-= (va
     assignExpressionTemplate (SubtractSingleFromVec<extent, decltype (self)> (self, c));
 }
 
+template <class ElementType, class StorageType, size_t extent, class StorageInfoType>
+ElementType VctrBase<ElementType, StorageType, extent, StorageInfoType>::min() const
+requires std::totally_ordered<ElementType>
```
pmglr, line 112: Do we need the `requires` both in the header and the implementation?

include/vctr/Containers/VctrBase.h
```diff
@@ -366,6 +366,24 @@ class VctrBase : private Config,
...
+    ElementType minAbs() const requires std::totally_ordered<ElementType>;
+
+    /** Returns the maximum value of all elements */
+    ElementType max() const requires std::totally_ordered<ElementType>;
+
+    /** Returns the maximum absolute value of all elements */
+    ElementType maxAbs() const requires std::totally_ordered<ElementType>;
+
+    /** Returns the sum of all elements */
+    ElementType sum() const requires has::operatorPlusEquals<ElementType>;
```
pmglr, line 402: I suggest (for the future) to also add the sum-of-squares, squared sum, and sum of the absolute value because I feel they are frequently needed. I'll check if there is an issue for that...

pmglr: Ah, I had even mentioned sum-of-squares in the issue description! Never mind, let's make a separate issue for that.

include/vctr/Containers/VctrBase.h
JanosGit: I'd say you need to implement these operators for your custom type in a way that they fulfil this requirement 🤷‍♂️

JanosGit: Aaah, I already tried to add a lot of them 😄 Will do!

include/vctr/Containers/VctrBase.tpp
JanosGit: Yes, the constraint becomes part of the function signature, just like all the templates. To me this is the biggest downside with concepts when declaration and definition of class templates are split 😕

I thought of wrapping all this in a macro but that would probably not make the code better understandable for others, so I'd stick to the verbose route here.

include/vctr/Containers/VctrBase.h
JanosGit: Good point, I think they are not on our feature backlog now, so feel free to add an issue

include/vctr/Containers/VctrBase.h
pmglr: I believe you should be able to locate all of them with this regex: `[^. ]{1} *\*/`

include/vctr/Containers/VctrBase.tpp
pmglr: I see, and I agree, that we should hide too much behind macros and over-specialized concepts that you have to look up before you can understand the function!

include/vctr/Containers/VctrBase.h
p4th4x0r: very interesting concept since you could also implement i.e. your own struct with overloaded += operator that does anything under the hood! nice choice.

include/vctr/Containers/VctrBase.tpp
p4th4x0r: hm, wouldn't it be better to add the definition also into the header for such short definitions? i mean it's just a one-liner with return statement.

JanosGit: This has to be defined here to break up cyclic inclusion dependency issues unfortunately 😔

include/vctr/Containers/VctrBase.h
JanosGit: Thanks! I try to implement generalised solutions like that wherever possible in the implementation

include/vctr/Expressions/ReductionExpression.h
```diff
@@ -0,0 +1,184 @@
...
+            if constexpr (is::floatNumber<ValueType<Expression>>)
+            {
+                if (Config::supportsAVX)
+                    return reduceAVX (e);
+            }
+            else
+            {
+                if (Config::supportsAVX2)
+                    return reduceAVX2 (e);
+            }
```
TheWaveWarden, line 70: In the logic above, why does the case `is::floatNumber` check for AVX and else checks for AVX2?

include/vctr/PlatformVectorOps/IntelIPP.h
```diff
@@ -110,6 +122,8 @@ class IntelIPP<std::complex<float>>
     static void mul (const std::complex<float>* srcA, std::complex<float> srcB,        std::complex<float>* dst, int len) { assertIppNoErr (ippsMulC_32fc (fc (srcA), fc (srcB), fc (dst), len)); }
     static void div (const std::complex<float>* srcA, const std::complex<float>* srcB, std::complex<float>* dst, int len) { assertIppNoErr (ippsDiv_32fc (fc (srcB), fc (srcA), fc (dst), len)); }
     static void div (const std::complex<float>* srcA, std::complex<float> srcB,        std::complex<float>* dst, int len) { assertIppNoErr (ippsDivC_32fc (fc (srcA), fc (srcB), fc (dst), len)); }
+
+    static std::complex<float> sum (const std::complex<float>* src, int len) { std::complex<float> r; assertIppNoErr (ippsSum_32fc (fc (src), len, fc (&r), ippAlgHintNone)); return r; }
```
TheWaveWarden, line 126: perhaps `maxAbs` and `minAbs` would also make sense for complex floats?

test/TestCases/Expressions/Max.cpp
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+#include <vctr_test_utils/vctr_test_common.h>
+
+TEMPLATE_PRODUCT_TEST_CASE ("Max", "[max]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double, int32_t, int64_t, uint32_t, uint64_t) )
+{
+    VCTR_TEST_DEFINES (10)
+
+    const auto max = vctr::max << filter << srcA;
+    const auto maxU = vctr::max << filter << srcUnaligned;
```
TheWaveWarden, line 30: You could add a src vector for a test case with negative only numbers :slightly_smiling_face:

TheWaveWarden (approved):
Very nice, looking forward to using these in my code :smiley_cat:

include/vctr/Expressions/Core/Max.h
JanosGit: You are so right, I already changed that in the meantime :)

include/vctr/Expressions/ReductionExpression.h
JanosGit: Because AVX2 supplies the integer instructions for AVX registers while AVX(1) supplies the floating point instructions

include/vctr/PlatformVectorOps/IntelIPP.h
JanosGit: they do from a logical point of view but those versions are not supplied by IPP unfortunately: https://www.intel.com/content/www/us/en/develop/documentation/ipp-dev-reference/top/volume-1-signal-and-data-processing/essential-functions/statistical-functions/maxabs.html

Since this class acts only as a thin abstraction wrapper to what is supplied by the IPP library it wouldn't make to add it here. But the expression implementation itself should work with complex floats

test/TestCases/Expressions/Max.cpp
JanosGit: Good point, will add that!

JanosGit: Has been added

PR #47, opened 2022-12-19
Vctr 45 implement find, find if

### LINKED ISSUE #1: Implement iterators

### LINKED ISSUE #21: Implement find, findIf

JanosGit:
I resolved https://github.com/sonible/VCTR/issues/1 along with this PR

### DIFF
--- a/include/vctr/Containers/VctrBase.h
+++ b/include/vctr/Containers/VctrBase.h
@@ -255,7 +255,7 @@ public:
             storage[i] = fn (storage[i], i);
     }
 
-    /** Calls a function on each element and passes the element index as second function argument.
+    /** Calls a function on each element and forwards fnArgs to the function after the value.
 
         The function returns void and might take the elements by reference to mutate them.
      */
@@ -266,7 +266,7 @@ public:
             fn (e, std::forward<Args> (fnArgs)...);
     }
 
-    /** Calls a function on each element and passes the element index as second function argument.
+    /** Calls a function on each element and forwards fnArgs to the function after the value.
 
         The function returns void and must not mutate the elements.
      */
@@ -277,7 +277,7 @@ public:
             fn (e, std::forward<Args> (fnArgs)...);
     }
 
-    /** Calls a function on each element and passes the element index as second function argument.
+    /** Calls a function on each element and forwards fnArgs to the function after the value.
 
         The function returns void and might take the elements by reference to mutate them.
      */
--- a/include/vctr/Containers/VctrBase.h
+++ b/include/vctr/Containers/VctrBase.h
@@ -310,15 +310,67 @@ public:
     //==============================================================================
     // Finding elements and manipulating them
     //==============================================================================
+    /** Returns an iterator to the first element that equals valueToLookFor or end() if none was found */
+    constexpr auto find (const ElementType& valueToLookFor)
+    {
+        return std::find (begin(), end(), valueToLookFor);
+    }
+
+    /** Returns a const iterator to the first element that equals valueToLookFor or end() if none was found */
+    constexpr auto find (const ElementType& valueToLookFor) const
+    {
+        return std::find (begin(), end(), valueToLookFor);
+    }
+
+    /** Returns a reverse iterator to the last element in this vector that equals p or rend() if none was found */
+    constexpr auto findReverse (const ElementType& valueToLookFor)
+    {
+        return std::find (rbegin(), rend(), valueToLookFor);
+    }
+
+    /** Returns a const reverse iterator to the last element in this vector that equals p or rend() if none was found */
+    constexpr auto findReverse (const ElementType& valueToLookFor) const
+    {
+        return std::find (rbegin(), rend(), valueToLookFor);
+    }
+
+    /** Returns an iterator to the first element in this vector for which pred returns true or end() if none was found */
+    template <is::functionWithSignatureOrImplicitlyConvertible<bool (const ElementType&)> Fn>
+    constexpr auto findIf (Fn&& predicate)
+    {
+        return std::find_if (begin(), end(), std::forward<Fn> (predicate));
+    }
+
+    /** Returns a const iterator to the first element in this vector for which pred returns true or end() if none was found */
+    template <is::functionWithSignatureOrImplicitlyConvertible<bool (const ElementType&)> Fn>
+    constexpr auto findIf (Fn&& predicate) const
+    {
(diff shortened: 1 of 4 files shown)


include/vctr/Containers/VctrBase.h
```diff
@@ -69,11 +69,30 @@ class VctrBase : private Config,
...
     static constexpr size_t size() requires (extent != std::dynamic_extent) { return extent; }
 
+    /** Returns an iterator to the begin of the storage */
     auto begin()       { return storage.begin(); }
+
+    /** Returns a const iterator to the begin of the storage */
     auto begin() const { return storage.begin(); }
+
+    /** Returns an iterator to the first element behind the storage */
     auto end()         { return is::stdArray<StorageType> ? storage.begin() + extent : storage.end(); }
```
iromur: Aha, so with concepts we finally have a constexpr ternary operator? Good to know…

include/vctr/Containers/VctrBase.h
```diff
@@ -259,7 +278,7 @@ class VctrBase : private Config,
```
iromur, line 278: This description (above) does not seem to fit the function?

include/vctr/Containers/VctrBase.h
```diff
@@ -270,7 +289,7 @@ class VctrBase : private Config,
```
iromur, line 289: Same here…

include/vctr/Containers/VctrBase.h
```diff
@@ -281,7 +300,7 @@ class VctrBase : private Config,
```
iromur, line 300: … and here.

include/vctr/Containers/VctrBase.h
```diff
@@ -281,7 +300,7 @@ class VctrBase : private Config,
 
         The function returns void and might take the elements by reference to mutate them.
```
iromur, line 301: … and also this one.

include/vctr/Containers/VctrBase.h
```diff
@@ -291,6 +310,58 @@ class VctrBase : private Config,
     //==============================================================================
     // Finding elements and manipulating them
     //==============================================================================
+    /** Returns an iterator to the first element that equaled valueToLookFor or end() if none was found */
```
iromur: I would prefer "element that equals" here and in the descriptions below as the simple past actually implies that it might have changed already, so it sounds a bit confusing to me (whereas "none was found" appears fine to me).

include/vctr/Containers/VctrBase.h
```diff
@@ -291,6 +310,58 @@ class VctrBase : private Config,
...
+        return std::find (rbegin(), rend(), valueToLookFor);
+    }
+
+    /** Returns a const reverse iterator to the last element in this vector that equaled p or rend() if none was found */
+    auto findReverse (const ElementType& valueToLookFor) const
+    {
+        return std::find (rbegin(), rend(), valueToLookFor);
+    }
+
+    /** Returns an iterator to the first element in this vector for which pred returned true or end() if none was found */
```
iromur: Similarly here, "returns".

include/vctr/Containers/VctrBase.h
```diff
@@ -291,6 +310,58 @@ class VctrBase : private Config,
...
+    template <is::functionWithSignatureOrImplicitlyConvertible<bool (const ElementType&)> F>
+    auto findIf (F&& pred)
+    {
+        return std::find_if (begin(), end(), pred);
```
iromur: So no need for `std::forward<F> (pred)` here because it is the only occurrence of the variable?

include/vctr/TypeTraitsAndConcepts/FunctionConcepts.h
```diff
@@ -20,36 +20,73 @@
...
+
+template <class, class>
+struct FunctionWithSignatureOrImplicitlyConvertible : std::false_type
+{
+};
+
+template <class Fn, class Ret, class... Args>
+struct FunctionWithSignatureOrImplicitlyConvertible<Fn, Ret (Args...)>
+{
+    static constexpr bool value = requires (Fn&& t, Args&&... args) { { t (args...) } -> std::convertible_to<Ret>; };
```
iromur, line 45: 🥇

iromur (approved):

include/vctr/Containers/VctrBase.h
JanosGit: Well, the ternary operator was allowed in a constexpr context since the beginning of constexpr afaik. This here is actually not like an `if constexpr` thing where both branches could return a different type, it's more like we can be sure that the compiler will figure out that the condition is compile time constant and so no branching will actually happen, so this whole thing looks a lot more compact

JanosGit: But your comment reminded me that these member functions are missing a constexpr keyword. Will add them...

include/vctr/Containers/VctrBase.h
JanosGit: I think the description might be a bit ambiguous if you only read the last line of the comment displayed here. The function that is meant here is the function object `fn` passed as an argument, which is expected to return something rather than that `forEach` member function. Does this become clearer to you if you read the entire comment or would you recommend rewording it?

include/vctr/Containers/VctrBase.h
JanosGit: Good point, will change that

include/vctr/Containers/VctrBase.h
JanosGit: Oh no, I missed forwarding it perfectly here. Thanks for the heads-up

include/vctr/Containers/VctrBase.h
iromur: Aha, I somehow remember that ternary op did not work in `constexpr` (unlike `if constexpr`) and that I also found a hint somewhere on stackxc, but maybe I am mixing up things.

include/vctr/Containers/VctrBase.h
iromur: No, I actually mean the first line of the comment (hidden here and not affected by your PR): "Calls a function on each element and passes the element index as second function argument." – But actually it does not pass the index but the `Args` provided in the parameter pack when calling `forEach()`.

JanosGit: Ah now I see. Yep, a copy & paste error. Will fix that :)

PR #48, opened 2022-12-22
VctrBase/Vector: Added resizing and size standard functions (VCTR-34, VCTR-35).

iromur:
This adds or reviews resize, reserve, shrink_to_fit, max_size, backIdx, empty, sizeInBytes, capacity, get_allocator.

### DIFF
--- a/include/vctr/Containers/VctrBase.h
+++ b/include/vctr/Containers/VctrBase.h
@@ -64,10 +64,37 @@ public:
     auto* data() const { return storage.data(); }
 
     /** Returns the number of elements. This overload is a non-static function, used in case the extent is dynamic. */
-    size_t size() const requires (extent == std::dynamic_extent) { return storage.size(); }
+    constexpr size_t size() const noexcept requires (extent == std::dynamic_extent) { return storage.size(); }
 
     /** Returns the number of elements. This overload is a static function, used in case the extent is static. */
-    static constexpr size_t size() requires (extent != std::dynamic_extent) { return extent; }
+    static constexpr size_t size() noexcept requires (extent != std::dynamic_extent) { return extent; }
+
+    /** Checks whether the container is empty. */
+    constexpr bool empty() const noexcept { return storage.empty(); }
+
+    /** Returns the container size in bytes.
+
+        This overload is a non-static function, used in case the extent is dynamic.
+     */
+    constexpr size_t sizeInBytes() const noexcept requires (extent == std::dynamic_extent) { return size() * sizeof (value_type); }
+
+    /** Returns the container size in bytes.
+
+        This overload is a static function, used in case the extent is static.
+     */
+    static constexpr size_t sizeInBytes() noexcept requires (extent != std::dynamic_extent) { return size() * sizeof (value_type); }
+
+    /** Returns the index referring to the last element in the vector.
+
+        This overload is a non-static function, used in case the extent is dynamic.
+     */
+    constexpr size_t backIdx() const noexcept requires (extent == std::dynamic_extent) { return size() - 1; }
+
+    /** Returns the index referring to the last element in the vector.
+
+        This overload is a static function, used in case the extent is static.
+     */
+    static constexpr size_t backIdx() noexcept requires (extent != std::dynamic_extent) { return size() - 1; }
 
     /** Returns an iterator to the begin of the storage */
     constexpr auto begin()       { return storage.begin(); }
--- a/include/vctr/Containers/Vector.h
+++ b/include/vctr/Containers/Vector.h
@@ -174,16 +174,66 @@ public:
         return { Vctr::begin(), Vctr::end() };
     }
 
+    /** Changes the size of this vector, potentially allocating memory.
+
+        This is a standard interface function forwarded to std::vector::resize().
+     */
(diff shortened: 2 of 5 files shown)


include/vctr/Containers/VctrBase.h
```diff
@@ -64,10 +64,49 @@ class VctrBase : private Config,
...
 
     /** Returns the number of elements. This overload is a static function, used in case the extent is static. */
-    static constexpr size_t size() requires (extent != std::dynamic_extent) { return extent; }
+    static constexpr size_t size() noexcept requires (extent != std::dynamic_extent) { return extent; }
+
+    /** Returns the maximum number of elements the container is able to hold.
+
+        This overload is a non-static function, used in case the extent is dynamic.
+      */
+    constexpr size_t max_size() const noexcept requires (extent == std::dynamic_extent) { return storage.max_size(); }
```
JanosGit: I think `max_size` is only specified in `std::vector`. So this code path would fail to compile for a `Span` with dynamic extent. [cppreference](https://en.cppreference.com/w/cpp/container/vector/max_size) says:

> Returns the maximum number of elements the container is able to hold due to system or library implementation limitations, i.e. [std::distance](http://en.cppreference.com/w/cpp/iterator/distance)(begin(), end()) for the largest container. [...] This value typically reflects the theoretical limit on the size of the container, at most [std::numeric_limits](http://en.cppreference.com/w/cpp/types/numeric_limits)<difference_type>::max().

So this is more about a theoretical system memory limitation for `std::vector` and probably rarely used in real world code. I'd propose to move this to the `Vector` implementation and drop the overload for the dynamic extent case

include/vctr/Containers/Vector.h
```diff
@@ -174,16 +174,30 @@ class Vector : public VctrBase<ElementType, std::vector<ElementType, Allocator<E
...
 
+    /** Standard interface function forwarded to std::vector. */
     template <class... Args>
     constexpr void emplace_back (Args&&... args) { Vctr::storage.emplace_back (std::forward<Args> (args)...); }
+
+    /** Standard interface function forwarded to std::vector. */
+    constexpr size_t capacity() const noexcept { return Vctr::storage.capacity(); }
+
+    /** Standard interface function forwarded to std::vector. */
+    constexpr Allocator<ElementType> get_allocator() const noexcept { return Vctr::storage.get_allocator(); }
```
JanosGit, line 233: I'm not sure if I prefer this comment style since it does not communicate directly what the functions are actually doing but requires the user to look up the `std::vector` documentation to find that out. I'd rather prefer e.g.

```
Returns the allocator associated with the Vector.

This is a standard interface function forwarded to std::vector::get_allocator.
```

include/vctr/Containers/Vector.h
```diff
@@ -174,16 +174,30 @@ class Vector : public VctrBase<ElementType, std::vector<ElementType, Allocator<E
...
     constexpr void reserve (size_t i) { Vctr::storage.reserve (i); }
 
+    /** Standard interface function forwarded to std::vector. */
+    constexpr void shrink_to_fit() { Vctr::storage.shrink_to_fit(); }
+
+    /** Standard interface function forwarded to std::vector. */
     constexpr void push_back (ElementType&& newElement) { Vctr::storage.push_back (std::move (newElement)); }
 
+    /** Standard interface function forwarded to std::vector. */
     constexpr void push_back (const ElementType& newElement) { Vctr::storage.push_back (newElement); }
```
JanosGit: The current `pago::Vector` implementation returns a reference to the last element instead of `void` which is quite handy in some situations. Could you add this feature here too?

test/TestCases/VctrBaseMemberFunctions.cpp
```diff
@@ -247,3 +247,19 @@ TEST_CASE ("forEach", "[VctrBaseMemberFunctions]")
...
+    vctr::Span spanFromV (v);
+    vctr::Span spanFromA (a);
+
+    constexpr auto sizeOfA = a.sizeInBytes();
+    constexpr auto sizeOfSpanFromA = spanFromA.sizeInBytes();
+
+    REQUIRE (v.sizeInBytes() == 400);
+    REQUIRE (spanFromV.sizeInBytes() == 400);
+    REQUIRE (sizeOfA == 400);
+    REQUIRE (sizeOfSpanFromA == std::span (a).size_bytes());
```
JanosGit, line 287: I'm a bit surprised to read `std::span (a)` here. Does this pick the single argument `range` constructor of `std::span`?

JanosGit:
Looks good overall!

Do you have any specific reasons to not add a unit test case for `empty`, `capacity` and `shrink_to_fit`?

test/TestCases/VctrBaseMemberFunctions.cpp
iromur: Good question… I would have expected that it picks the single-argument `array` constructor? I just wanted to check against `std::span::size_bytes()`, which is seems to be implemented only there.

include/vctr/Containers/Vector.h
iromur: Added. Implementation could be probably changed to back() in your task.

JanosGit (approved):
All good now, thank you 👍

PR #53, opened 2022-12-28
Vctr 39 implement erase, insert, append, release, swap

### LINKED ISSUE #18: Implement erase, insert, append, release, swap

### LINKED ISSUE #26: Implement erase, eraseIf, eraseAll, eraseAllIf

JanosGit:


include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+    // Erasing elements from the Vector
+    //==============================================================================
+    /** Erases the element referenced by elementToErase and returns the iterator to the element behind it.
+
+        This is a standard interface function forwarded to std::vector::erase().
+     */
+    constexpr auto erase (ConstIterator elementToErase)
+    {
+        // Undefined behaviour if elementToErase does not refer to an element in this Vector
+        VCTR_ASSERT (Vctr::contains (elementToErase));
```
p4th4x0r, line 287: hm, yes maybe there would be undefined behaviour if the element could not be found. however, std::erase does nothing imo when the element could not be found. so why don't you add an if (contains) instead of assert here?

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+        return Vctr::storage.erase (elementToErase);
+    }
+
+    /** Erases the element at index idx and returns the iterator to the element behind it. */
+    constexpr auto erase (size_t idx)
+    {
+        return Vctr::storage.erase (Vctr::storage.begin() + idx);
```
p4th4x0r: what if the index exceeds bounds?

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+        VCTR_ASSERT (Vctr::contains (last) || last == Vctr::end());
+
+        return Vctr::storage.erase (first, last);
+    }
+
+    /** Erases numElements elements starting from index startIdx and returns the iterator to the element behind it. */
+    constexpr auto erase (size_t startIdx, size_t numElements)
+    {
+        auto it = Vctr::storage.begin() + startIdx;
+        return Vctr::storage.erase (it, it + numElements);
```
p4th4x0r: what if numElements or startIdx or startIdx + numElements exceeds limits?

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+    /** Erases the first occurrence of value from this Vector and adjusts its size. */
+    template <std::equality_comparable_with<ElementType> T>
+    constexpr auto eraseFirstOccurrenceOf (const T& value)
+    {
+        return erase (Vctr::find (value));
+    }
+
+    /** Erases the first occurrence for which predicate is true from this Vector and adjusts its size. */
+    template <is::functionWithSignatureOrImplicitlyConvertible<bool (const ElementType&)> Fn>
+    constexpr auto eraseFirstIf (Fn&& predicate)
```
p4th4x0r: hm, the name is a bit misleasing. maybe someone will understand this as: "remove the **first element** of the vector if the predicate returns true"

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+    constexpr void eraseAllIf (Fn&& predicate)
+    {
+        erase (std::remove_if (Vctr::begin(), Vctr::end(), std::forward<Fn> (predicate)), Vctr::end());
+    }
+
+    //==============================================================================
+    // Adding elements to the Vector
+    //==============================================================================
+
+    /** Appends a Span, Array or Vector to the end of this Vector, optionally by moving elements from the source */
```
p4th4x0r, line 354: what does that mean by "moving elements from the source"?? moving to where? o0

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+        // Undefined behaviour if pos does not refer to an element in this Vector
+        VCTR_ASSERT (Vctr::contains (pos) || pos == Vctr::end());
+        return Vctr::storage.insert (pos, value);
+    }
+
+    /** Inserts value at index idx and returns an iterator to the inserted value. */
+    auto insert (size_t idx, const ElementType& value)
+    {
+        VCTR_ASSERT (idx < Vctr::size());
+        return Vctr::storage.insert (Vctr::begin() + idx, value);
```
p4th4x0r: could also just be forwarded to function in 365 `insert (Vctr::begin() + idx, value);`

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+        // Undefined behaviour if pos does not refer to an element in this Vector
+        VCTR_ASSERT (Vctr::contains (pos) || pos == Vctr::end());
+        return Vctr::storage.insert (pos, std::move (value));
+    }
+
+    /** Inserts value at index idx and returns an iterator to the inserted value. */
+    auto insert (size_t idx, ElementType&& value)
+    {
+        VCTR_ASSERT (idx < Vctr::size());
+        return Vctr::storage.insert (Vctr::begin() + idx, std::move (value));
```
p4th4x0r: also here maybe just forward to upper insert function?

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+    }
+
+    /** Inserts count copies of value before the element referenced by pos and returns an iterator to the
```
p4th4x0r: does "inserts count copies" mean that it copies the element that much times? if yes, i'd rather name it numCopies or somethimg similar.

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+
+    /** Inserts a list of values at index idx and returns an iterator to the first inserted value. */
+    auto insert (size_t idx, std::initializer_list<ElementType> initList)
+    {
+        VCTR_ASSERT (idx < Vctr::size());
+        return Vctr::storage.insert (Vctr::begin() + idx, initList);
+    }
+
+    /** Inserts a VCTR container before the element referenced by pos and returns an iterator to the
+        first inserted value.
```
p4th4x0r, line 475: moveValuesFromSrc is not documented here.

include/vctr/Containers/Vector.h
p4th4x0r: ah okay so the source vector moves into that vector and doesn't exist anymore?

include/vctr/Containers/Vector.h
```diff
@@ -238,30 +243,284 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+
+        // If it is no view and no reference, the container passed in has no dependencies outside this scope,
+        // so we can safely move its elements in every case
+        if constexpr ((! is::view<VctrToInsert>) && (! std::is_reference_v<VctrToInsert>))
+            moveValuesFromSrc = true;
+
+        if (moveValuesFromSrc)
+            return Vctr::storage.insert (Vctr::begin() + idx, std::make_move_iterator (vctrToInsert.begin()), std::make_move_iterator (vctrToInsert.end()));
+
+        return Vctr::storage.insert (Vctr::begin() + idx, vctrToInsert.begin(), vctrToInsert.end());
```
p4th4x0r: here the forwarding would be pretty nice huh by just writing:
`VCTR_ASSERT (idx < Vctr::size());`
`insert (Vctr::begin() + idx, vctrToInsert, moveValuesFromSrc);`

p4th4x0r (changes requested):
i don't like github

include/vctr/Containers/Vector.h
JanosGit: An iterator that does not point to any element in this container will probably come from another container. Passing an iterator from another container to erase is most likely a programming error that I'd like the programer to inform about here.

include/vctr/Containers/Vector.h
JanosGit: Good point, I'll add a bounds check

include/vctr/Containers/Vector.h
JanosGit: I see your point. Do you have a better name suggestion?

include/vctr/Containers/Vector.h
JanosGit: Yes, there might be cases where you want to reuse the source vector after this and copy the values when appending them. In other cases you might know that the elements held by the vector are no longer needed, then moving them out of the source vector into the destination can be a viable option

include/vctr/Containers/Vector.h
JanosGit: I'd now go with `eraseFirstOccurrenceIf` aligning with `eraseFirstOccurrenceOf`

include/vctr/Containers/Vector.h
JanosGit: done!

include/vctr/Containers/Vector.h
JanosGit: added!

include/vctr/Containers/Vector.h
JanosGit: I renamed it. The original name was taken from the standard lib implementation

include/vctr/Containers/Vector.h
JanosGit: yes!

include/vctr/Containers/Vector.h
JanosGit: yes!

include/vctr/Containers/Vector.h
JanosGit: This will now assert

include/vctr/Containers/Vector.h
p4th4x0r: Sounds perfect!

include/vctr/Containers/Vector.h
p4th4x0r: Sure! overlooked that you pass an iterator here! everything fine :)

p4th4x0r (approved):
Nice one, except the small requested changes in signature :)

JanosGit:
> Nice one, except the small requested changes in signature :)

You mean `eraseFirstIf` to `eraseFirstOccurrenceIf`? If so, this has already been changed with the latest push. Or do you refer to a different request that I overlooked?

PR #58, opened 2023-01-19
Vctr 88 log10.h gcem log10 doesn t compile

JanosGit:
This should fix the errors you were encountering. `gcem::log10` was introduced with version 1.16.0, these changes will warn if the installed version is below that.


p4th4x0r (approved):
Nothing to complain!

PR #60, opened 2023-01-23
Implement fill and init

### LINKED ISSUE #6: Implement fillRamp, fillSlope, write unit tests for init and fill

pmglr:


include/vctr/Containers/Vector.h
```diff
@@ -225,6 +225,19 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
      */
     constexpr void shrink_to_fit() { Vctr::storage.shrink_to_fit(); }
 
+    /** Fills the allocated memory with the given value. */
```
JanosGit: Referring to "allocated memory" here seems a bit misleading – I mean the way vectors are implemented this is likely to not be true since most of the time more memory is allocated than needed and not all of this memory will be filled by this function.

I'd simply say "fills all elements of this vector"

JanosGit (approved):
Minor documentation remark, otherwise all good!

include/vctr/Containers/Vector.h
pmglr: Good point.

JanosGit (approved):

PR #61, opened 2023-01-24
VctrBase: move fill and init here, write tests

pmglr:


pmglr:
@JanosGit I needed to access fill in vctr::Array, I hope VctrBase is the right place for the implementation. 

Probably I should also restrict the value type like copyFrom does?

include/vctr/Containers/VctrBase.h
```diff
@@ -247,6 +247,28 @@ class VctrBase : private Config,
         std::copy_n (otherData, otherSize, begin());
     }
 
+    /** Fills the container with the given value. */
+    constexpr void fill (const value_type& value) {  std::fill (storage.begin(), storage.end(), value); }
```
JanosGit: Moving fill to base is absolutely correct since this should be available to all container types. Please note however that `storage.begin()` and `storage.end()` might fill more than needed as we are using overly large arrays under the hood to have memory regions that are always a multiple of a simd register size. In this particular case, the worst thing that can happen is that we write to unused dummy memory, which is safe, the best thing that can happen is that the compiler makes use of this over-allocated memory and uses some simd acceleration to speed this up. Not sure about the cleverness of the compiler though ;) 

TLDR: Usually, you should always use `begin()` and `end()` in VctrBase instead of `storage.begin()` and `storage.end()` except you really know what you are doing. In this particular case this is okay

include/vctr/Containers/VctrBase.h
```diff
@@ -247,6 +247,28 @@ class VctrBase : private Config,
...
+
+    /** Resizes this container to newSize and fills it with the given value.
+
+        @see resize, fill.
+     */
+    constexpr void init (size_t newSize, const value_type& value) { storage.resize (newSize); fill (value);  }
```
JanosGit: In contrast to fill, init should only be available to `vctr::Vector` since it relies on the container having a resize function which is only the case for `vctr::Vector`. So please move the two init functions back to that class.

JanosGit (changes requested):
See notes in the code. `init` only makes sense in `vctr::Vector`

test/TestCases/VctrBaseMemberFunctions.cpp
```diff
@@ -352,3 +352,37 @@ TEST_CASE ("shift", "[VctrBaseMemberFunctions]")
...
+TEST_CASE ("init", "[VctrBaseMemberFunctions]")
+{
+    constexpr int fillValue = 13;
+
+    SECTION ("Reduce size")
+    {
+        vctr::Vector b { 0, 1, 2, 3, 4 };
+        b.init (3, fillValue);
+        REQUIRE (b.size() == 3);
+        REQUIRE (b.sum() == fillValue * b.size());
```
JanosGit: I'd rather go for `REQUIRE (b.allElementsEqual (fillValue))`

pmglr: Thx, didn't know about the method.

include/vctr/Containers/VctrBase.h
pmglr: Thanks for this very useful comment.

pmglr: Using begin/end now.

test/TestCases/VctrBaseMemberFunctions.cpp
```diff
@@ -352,3 +352,37 @@ TEST_CASE ("shift", "[VctrBaseMemberFunctions]")
...
     }
 }
+
+TEST_CASE ("fill", "[VctrBaseMemberFunctions]")
+{
+    constexpr int fillValue = 5;
+
+    vctr::Array a { 0, 1, 2, 3, 4 };
+    a.fill (fillValue);
+    REQUIRE (a.allElementsEqual(fillValue));
```
pmglr: Forgot to clang-format...

pmglr: done

JanosGit (approved):
Looks good now. Thanks for you contributions :)

PR #62, opened 2023-02-08
Vctr 62 implement rms mean square mean

JanosGit:


include/vctr/Expressions/Core/Mean.h
```diff
@@ -0,0 +1,243 @@
...
+    requires is::suitableForAccelerateRealFloatVectorReductionOp<SrcType, value_type, detail::dontPreferIfIppAndAccelerateAreAvailable>
+    {
+        return Expression::Accelerate::mean (src.data(), size());
+    };
+
+    VCTR_FORCEDINLINE value_type reduceVectorOp() const
+    requires is::suitableForIppRealOrComplexFloatVectorReductionOp<SrcType, value_type, detail::preferIfIppAndAccelerateAreAvailable>
+    {
+        return Expression::IPP::mean (src.data(), sizeToInt (size()));
+    };
```
pmglr, line 52: Is it on puprose that some of the reduce methods compute the sum and others compute the mean?

pmglr: Is finalizeReduction not called in this case?

JanosGit: Well, there are two ways this expression can be used. Either `reduceVectorOp` is called once to retrieve the entire result in one. Or `reduceElementWise`/`reduce[SIMD]RegisterWise` is called in a loop over all elements and the final result of this is passed to `finalizeReduction` where final computation steps are applied. See Expressions/Readme.md for more details.

In case of a mean computation the element wise calls indeed only sum up and the finalize calls performs the division of the sum.

test/TestCases/Expressions/Mean.cpp
```diff
@@ -0,0 +1,137 @@
...
+    GNU Lesser General Public License version 3 for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+#include <vctr_test_utils/vctr_test_common.h>
+
+TEMPLATE_PRODUCT_TEST_CASE ("Mean", "[mean]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double, std::complex<float>, std::complex<double>, int32_t, int64_t, uint32_t, uint64_t) )
```
pmglr, line 25: If we want to make sure that all the different SIMD implementations yield the correct results, I assume we need to run these test cases on the respective architectures. Do you know if all are covered? 🤔

JanosGit: Yes, they are hidden inside the `VCTR_NATIVE_SIMD` macro. The catch `TEMPLATE_PRODUCT_TEST_CASE` will then create a test case for every architecture in the list for every type in the list, kind of a test case matrix :)

include/vctr/Expressions/Core/Mean.h
pmglr: I see! Admittedly, didn't know how exactly it works!

test/TestCases/Expressions/Mean.cpp
```diff
@@ -0,0 +1,137 @@
...
+        REQUIRE_THAT (rms, Catch::Matchers::WithinRel (ref, eps));
+        REQUIRE_THAT (rmsU, Catch::Matchers::WithinRel (refU, eps));
+    }
+    else
+    {
+        REQUIRE (rms == rmsMemFn);
+        REQUIRE (rms == ref);
+        REQUIRE (rmsU == refU);
+    }
+}
```
pmglr, line 137: Very tidy test cases

include/vctr/PlatformVectorOps/IntelIPP.h
```diff
@@ -98,11 +101,13 @@ class IntelIPP<double>
...
-    static double minAbs (const double* src, int len) { double r; assertIppNoErr (ippsMinAbs_64f (src, len, &r)); return r; }
-    static double sum    (const double* src, int len) { double r; assertIppNoErr (ippsSum_64f (src, len, &r)); return r; }
+    static double max        (const double* src, int len) { double r; assertIppNoErr (ippsMax_64f (src, len, &r)); return r; }
+    static double maxAbs     (const double* src, int len) { double r; assertIppNoErr (ippsMaxAbs_64f (src, len, &r)); return r; }
+    static double min        (const double* src, int len) { double r; assertIppNoErr (ippsMin_64f (src, len, &r)); return r; }
+    static double minAbs     (const double* src, int len) { double r; assertIppNoErr (ippsMinAbs_64f (src, len, &r)); return r; }
+    static double sum        (const double* src, int len) { double r; assertIppNoErr (ippsSum_64f (src, len, &r)); return r; }
+    static double mean       (const double* src, int len) { double r; assertIppNoErr (ippsMean_64f (src, len, &r)); return r;}
+    static double meanSquare (const double* src, int len) { double r; assertIppNoErr (ippsNorm_L2_64f (src, len, &r)); return r * r;}
```
pmglr: Looks like there was some research involved in finding all the right intrinsics!

pmglr (approved):

include/vctr/PlatformVectorOps/IntelIPP.h
JanosGit: The [Intel IPP Developer Reference](https://www.intel.com/content/www/us/en/develop/documentation/ipp-dev-reference/top.html) and [Apple Accelerate Documentation](https://developer.apple.com/documentation/accelerate?language=objc) are my go-to resources for that, they make identifying the right more or less cryptic function names a bit easier 😉

But turns out that I introduced an error while trying to (ab)use Norm_L2 for meanSquare, which led to the unit test fails. Should be fixed now!

PR #66, opened 2023-02-24
Generators: Implement Linspace (aka ramp/slope)

### LINKED ISSUE #12: Implement ramp and slope

pmglr:


include/vctr/Containers/VctrBase.h
```diff
@@ -35,16 +35,16 @@ class VctrBase : private Config,
...
 
         In case the instance specifies a dynamic extent, the return value will always be
         std::dynamic_extent, no matter if we specified an amount to shrink.
         In case the instance specifies a static extent, the return value will be extent - amountToShrink.
      */
-    static consteval size_t getExtent (size_t amountToShrink = 0)
-    {
-        return extent == std::dynamic_extent ? std::dynamic_extent : (extent - amountToShrink);
-    }
+    static consteval size_t getExtent (size_t amountToShrink = 0) {  return extent == std::dynamic_extent ? std::dynamic_extent : (extent - amountToShrink);  }
```
JanosGit: Why this change?

include/vctr/Containers/VctrBase.h
```diff
@@ -251,7 +253,42 @@ class VctrBase : private Config,
...
-    constexpr void fill (const value_type& value) {  std::fill (begin(), end(), value); }
+    constexpr void fill (const value_type& value) { std::fill (begin(), end(), value); }
+
+    //==============================================================================
+    // Generators
+    //==============================================================================
+    /** Fills the vector with evenly spaced numbers between start and stop.
+
+        The length of the vector will not be changed and must be > 0.
+        If start and stop are identical, the value will be filled with that value.
```
JanosGit: Probably you meant something like `If start and stop are identical, all elements will be filled with that value.`?

include/vctr/Containers/VctrBase.h
```diff
@@ -251,7 +253,42 @@ class VctrBase : private Config,
...
+
+        @param start      First value of the interval; always included.
+        @param stop       Last value of the interval; only included if includeEnd is true.
+        @param includeEnd If this is true, the numbers lie on the closed interval
+                          [start, stop]; otherwise the interval is half-open [start, stop).
+     */
+    void fillLinspace (ElementType start, ElementType stop, bool includeEnd = true)
+    requires is::number<ElementType>
```
JanosGit: `is::number` also includes complex numbers, in the unit test you are only testing with real values. I'm wondering, is this well defined for complex numbers? If so, please add tests for complex numbers, if not better constrain it to `is::realNumber`

include/vctr/Containers/VctrBase.h
```diff
@@ -251,7 +253,42 @@ class VctrBase : private Config,
...
+
+        VCTR_ASSERT (num > 0);
+        VCTR_ASSERT (! includeEnd || num != 1); // a num-one range cannot include the end
+
+        const auto increment = (stop - start) / double (num - int (includeEnd));
+
+        const auto isIntegerIncrement = increment - std::floor (increment) == ElementType (0);
+        VCTR_ASSERT (std::is_floating_point_v<ElementType> || isIntegerIncrement);
+
+        auto value = double (start);
```
JanosGit, line 289: Wait, this is definitely not well defined for complex numbers ;)

include/vctr/Containers/VctrBase.h
```diff
@@ -251,7 +253,42 @@ class VctrBase : private Config,
...
+     */
+    void fillLinspace (ElementType start, ElementType stop, bool includeEnd = true)
+    requires is::number<ElementType>
+    {
+        const auto num = size();
+
+        VCTR_ASSERT (num > 0);
+        VCTR_ASSERT (! includeEnd || num != 1); // a num-one range cannot include the end
+
+        const auto increment = (stop - start) / double (num - int (includeEnd));
```
JanosGit: I'm not sure if subtracting an int from a size_t could result in compiler warnings in some situations, maybe better cast `includeEnd` to `size_t`

include/vctr/Containers/VctrBase.h
```diff
@@ -251,7 +253,42 @@ class VctrBase : private Config,
...
+        @param includeEnd If this is true, the numbers lie on the closed interval
+                          [start, stop]; otherwise the interval is half-open [start, stop).
+     */
+    void fillLinspace (ElementType start, ElementType stop, bool includeEnd = true)
+    requires is::number<ElementType>
+    {
+        const auto num = size();
+
+        VCTR_ASSERT (num > 0);
+        VCTR_ASSERT (! includeEnd || num != 1); // a num-one range cannot include the end
```
JanosGit, line 282: Should we assert that `start < stop`?

include/vctr/Generators/Linspace.h
```diff
@@ -0,0 +1,82 @@
...
+
+       @tparam ElementType Element type contained in the vector.
+
+       @param start        First value of the interval; always included.
+       @param stop         Last value of the interval; only included if includeEnd is true.
+       @param num          Number of elements (length of the vector), must be > 0.
+       @param includeEnd   If this is true, the numbers lie on the closed interval
+                           [start, stop]; otherwise the interval is half-open [start, stop).
+     */
+    template <class ElementType>
```
JanosGit: I'd constrain ElementType to e.g. `is::realNumber` here and below

test/TestCases/Generators/Linspace.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+        vctr::Vector<TestType> result (num);
+        result.fillLinspace (value, value, true);
+
+        REQUIRE_THAT (result, vctr::Equals (vctr::generators::Linspace::makeVector (value, value, num, true)));
+
+        // value should be repeated:
+        // >>> import numpy as np
+        // >>> np.linspace(start=3.0,stop=3.0, num=10, endpoint=True)
+        // array([3., 3., 3., 3., 3., 3., 3., 3., 3., 3.])
```
JanosGit, line 56: Is this piece of python code a leftover? ;)

test/TestCases/Generators/Linspace.cpp
```diff
@@ -0,0 +1,197 @@
...
+        REQUIRE_THAT (result, vctr::Equals (vctr::Vector (result.size(), value)));
```
JanosGit: `REQUIRE (result.allElementsEqual (value));` would be even more expressive in my opinion

test/TestCases/Generators/Linspace.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+                REQUIRE (withEnd.back() == stop);
+                REQUIRE (withoutEnd.back() == stop + pos * TestType (1));
+                REQUIRE_THAT (withEnd.subSpan (0, 10), vctr::Equals (withoutEnd));
+            }
+
+            SECTION ("integer step reverse")
+            {
+                // >>> import numpy as np
+                // >>> np.linspace(start=3.0, stop=-3.0, num=7, endpoint=True)
```
JanosGit: Again some python here 👀

test/TestCases/Generators/Linspace.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+        auto vector = vctr::generators::Linspace::makeVector<TestType> (start, stop, num, includeEnd);
+        auto array = vctr::generators::Linspace::makeArray<TestType, num> (start, stop, includeEnd);
+
+        REQUIRE (vector.size() == array.size());
+
+        REQUIRE_THAT (vector, vctr::Equals (array));
+    }
+
+    for (auto sign : { TestType (1), TestType (-1) })
```
JanosGit: This kind of loops can better be replaced with a GENERATE statement

test/TestCases/Generators/Linspace.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+                for (auto n = 0; n < num; ++n)
+                    REQUIRE (reference[n] == Approx (result[n]));
+            }
+
+            SECTION ("Precision")
+            {
+                TestType start = pos * TestType (0.73);
+                TestType stop = pos * TestType (1);
+                size_t num = 123456789;
```
JanosGit, line 188: a solid size choice 😄

include/vctr/Containers/VctrBase.h
pmglr: A lot of the methods in this file were not clang-format safe. This is why I added the guards. This one I aligned to the style of the other methods that are single-line in majority.

include/vctr/Containers/VctrBase.h
pmglr: Thx!

include/vctr/Containers/VctrBase.h
pmglr: Good point!

include/vctr/Containers/VctrBase.h
pmglr: No. If you have a look at the tests, you'll see that we can go in both directions.

include/vctr/Containers/VctrBase.h
JanosGit: I see... this one is quite long for a single liner in my opinion. Maybe we should even go and move all non-single liners (or really all definitions) to the .tpp file?

include/vctr/Containers/VctrBase.h
JanosGit: I think this is also found a few times in the `generators::Linspace` class

include/vctr/Containers/VctrBase.h
JanosGit: Ah right, wrote this before going through the test in detail.

JanosGit: I wonder if we should explicitly mention this feature in the documentation?

test/TestCases/Generators/Linspace.cpp
pmglr: This is the test baseline

JanosGit: I see, makes sense. Maybe add a general comment above the test case that this tries to check against the results of the python code found in the test?

include/vctr/Containers/VctrBase.h
pmglr: I don't know - my main concern was here that the file wasn't clang-format safe. This might be a long line, but seems tolerable - in contrast to many others that have _multiple_ statements within one line, in this fashion:

```
type function() { VCTR_ASSERT (something); return somethingElse; }
```

pmglr: But I can revert this particular change for getExtent if you like

JanosGit: No leave it as it is, I'd rather add a general cleanup task for this file to the backlog

include/vctr/Containers/VctrBase.h
pmglr: Sure, why not :)

pmglr: Than also means it's not correct to define the interval as [start, stop) which would only be true if start < stop

test/TestCases/Generators/Linspace.cpp
pmglr: I added comments like this one: `// value should be repeated, compare numpy's linspace:`

include/vctr/Containers/VctrBase.h
pmglr: Now it's too late :)

test/TestCases/Generators/Linspace.cpp
pmglr: Hehe

JanosGit (approved):

PR #68, opened 2023-02-24
Vctr 52 implement multiply and add

### LINKED ISSUE #27: Implement multiplyAndAdd

JanosGit:


test/TestCases/Expressions/Add.cpp
```diff
@@ -54,4 +54,36 @@ TEMPLATE_PRODUCT_TEST_CASE ("Addition in place", "[add]", (PlatformVectorOps, VC
...
     REQUIRE_THAT (sum2, vctr::EqualsTransformedBy<addition> (srcA, c).withEpsilon());
+
+    // Operator += implements special accelerated handling for the case of multiply-accumulate operations
+    auto mac1 = srcA;
+    auto mac2 = srcA;
+    //auto mac3 = srcA;
+
+    mac1 += srcB * srcC;
+    mac2 += srcB * c;
+    //mac3 += vctr::multiplyByConstant<42> << srcB;
```
JanosGit, line 66: This one did not build due to some compiler errors out of the scope of this task. As it's quite exotic I'll rather add a task to the backlog to fix that. Issue link is coming soon... ;)

pmglr:
Is it correct that you refer to multiply-accumulate in a situation like this

```
a = a + b * c /* or */ a += a * b
```

and to multiply-add in the following case?

```
d = a + b * c
```

include/vctr/Expressions/Core/Multiply.h
```diff
@@ -78,6 +78,19 @@ class MultiplyVectors : ExpressionTemplateBase
...
 
+
+    VCTR_FORCEDINLINE void evalVectorOpMultiplyAccumulate (value_type* dst) const
+    requires is::suitableForAccelerateRealFloatBinaryVectorOp<SrcAType, SrcBType, value_type, detail::dontPreferIfIppAndAccelerateAreAvailable> && is::anyVctr<SrcAType> && is::anyVctr<SrcBType>
+    {
+        Expression::Accelerate::multiplyAdd (srcA.data(), srcB.data(), dst, dst, size());
+    }
+
+    VCTR_FORCEDINLINE void evalVectorOpMultiplyAccumulate (value_type* dst) const
+    requires is::suitableForIppRealOrComplexComplexFloatBinaryVectorOp<SrcAType, SrcBType, value_type, detail::preferIfIppAndAccelerateAreAvailable> && is::anyVctr<SrcAType> && is::anyVctr<SrcBType>
```
pmglr, line 89: `suitableForIppRealOrComplexComplexFloatBinaryVectorOp` Can we come up with a longer one? 😅

pmglr, line 89: Is this `preferIfIppAndAccelerateAreAvailable` because `IPP::multiplyAccumulate` is faster than `Accelerate::multiplyAdd` (that has an extra destination parameter and presumably includes an extra copy step because of that)?

test/TestCases/Expressions/Add.cpp
pmglr, line 66: Is the `+=` operator the only way to access a multiply-accumulate operation of vctr?

JanosGit:
> Is it correct that you refer to multiply-accumulate in a situation like this
> 
> ```
> a = a + b * c /* or */ a += a * b
> ```
> 
> and to multiply-add in the following case?
> 
> ```
> d = a + b * c
> ```

Yes, at least when it comes to the platform accelerated vector operations. This was the naming convention that I extracted from the documentation of IPP and Accelerate

include/vctr/Expressions/Core/Multiply.h
JanosGit: Hahaha, you can't imagine how long I thought about the shortest possible names that communicate what these concepts do. 🤯

JanosGit: Yes, my limited set of benchmarks revealed that IPP outperforms Accelerate in nearly all cases. This is why I default for IPP in all cases, except for some rare cases where Accelerate provides a straightforward alternative for something that needs multiple chained IPP calls.

As the current goal is feature completeness, I'd go with this pattern for now. Once we are done implementing all features, I'd love to create some benchmarks and then review the choices again based on the outcome of these tests

test/TestCases/Expressions/Add.cpp
JanosGit: Well, you could theoretically also go for
```C++
(srcB * srcC).evalVectorOpMultiplyAccumulate (dstPointer);
```
or
```C++
vctr::Expressions::Multiply<extent> (srcB, srcC).evalVectorOpMultiplyAccumulate (dstPointer);
```
But I don't see why you would want that 😉

pmglr: Wouldn't  it make sense also to do that in the test to _really_ make sure that, in the test,

```
 v.evalVectorOpMultiplyAccumulate (data());
```

is executed and not

```
assignExpressionTemplate (Expressions::AddVectors<extent, decltype (self), const V&> (self, v));
```
?

pmglr:
Just let me know when the PR is in it's final state ;)

test/TestCases/Expressions/Add.cpp
JanosGit: It would not really make sense to me since that is an implementation detail. It's not in the scope of this test to check which variation to compute the result is used but to verify that the computation result is correct. Depending on the environment, `evalVectorOpMultiplyAccumulate` might or might not exist, e.g. for all non-apple builds without IPP this might be the case. And e.g. in case IPP is available this function is enabled for complex floats while it is only available for real floats in case only Accelerate can be used.

TLDR: I usually ensure that the dispatching works locally by setting some breakpoints before raising the PR, so I'm pretty confident that this works. The worst failure would be a slightly slower execution

JanosGit:
> Just let me know when the PR is in it's final state ;)

It is now. Had to fix some semi-related issues with the custom allocators, but now it works. I'll just squash some commits now, but code is ready for review

include/vctr/Expressions/Core/Multiply.h
p4th4x0r: however, "OrComplexComplex"? is there one "complex" too much? :)

test/TestCases/Expressions/Abs.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include <vctr_test_utils/vctr_test_common.h>
 
-TEMPLATE_PRODUCT_TEST_CASE ("Abs", "[abs]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double, int32_t, int64_t, std::complex<float>, std::complex<double>, std::complex<int32_t>, std::complex<int64_t>) )
```
p4th4x0r, line 25: why remove?

p4th4x0r (approved):

include/vctr/Expressions/Core/Multiply.h
JanosGit: Nope that's intentional. It converts a complex source into a complex destination. For e.g. abs this would be `RealOrComplexReal`. As I said I tried to keep a balance between verbosity and length. But I'm open for renaming proposals as long as they result in a consistent pattern across all those concepts

test/TestCases/Expressions/Abs.cpp
JanosGit: Mhm, the diff looks like this, not sure why GitHub does not display the additions..?
image

pmglr (approved):

test/TestCases/Expressions/Abs.cpp
JanosGit: In case your question was why I removed the `std::complex<int>` types: I had to find out that `std::complex` is only defined for floating point types and causes build errors on some platforms if used with integer types.

PR #72, opened 2023-02-28
Test Utils: add Test for Equals matcher

pmglr:
This test fails for double. I suspected that this test should pass for both double and float. @JanosGit @iromur do you have an idea why?

### DIFF
--- a/test/TestCases/TestUtils.cpp
+++ b/test/TestCases/TestUtils.cpp
@@ -1,0 +1,68 @@
+/*
+  ==============================================================================
+    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+
+    Copyright 2022- by sonible GmbH.
+
+    This file is part of VCTR - Versatile Container Templates Reconceptualized.
+
+    VCTR is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License version 3
+    only, as published by the Free Software Foundation.
+
+    VCTR is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License version 3 for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+#include <vctr_test_utils/vctr_catch_matchers.h>
+#include <vctr_test_utils/vctr_test_common.h>
+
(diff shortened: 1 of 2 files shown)


test/TestCases/TestUtils.cpp
```diff
@@ -0,0 +1,68 @@
...
+        const vctr::Vector<TestType> reference (10, refValue);
+
+        vctr::Vector<TestType> v (10, refValue);
+        REQUIRE_THAT (v, vctr::Equals (reference));
+
+        v.fill (testValue);
+
+        // margin
+        auto marginPass = std::abs (testValue - refValue);
+        auto marginFail = marginPass - std::numeric_limits<TestType>::epsilon();
```
iromur, line 42: Hmm, shouldn't read marginPass + epsilon? And is it guaranteed that, when you subtract, marginFail does not get smaller than 0?

iromur (commented):
Which catch version do we have?
I found this: https://stackoverflow.com/questions/64121152/how-to-compare-floating-point-in-catch2
Which recommends using Catch::Matchers when comparing floating point with catch >= 2.10.

p4th4x0r (approved):

p4th4x0r (approved):

test/TestCases/TestUtils.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+        // baseline test
+        for (auto n = 0; n < reference.size(); ++n)
+        {
+            CHECK_THAT (vWithin[n], Catch::Matchers::WithinRel (reference[n], epsilon));
+            CHECK_FALSE (Catch::Matchers::WithinRel (reference[n], epsilon).match (vOutside[n]));
+        }
+
+        REQUIRE_THAT (vWithin, vctr::Equals (reference).withEpsilon (epsilon));
+        REQUIRE_FALSE (vctr::Equals (reference).withEpsilon (epsilon).match (vOutside));
```
pmglr, line 63: I updated the test. Now we have one remaining fail which seems to reveal a difference between the matchers of catch and vctr:

```
/Users/maxgiller/Development/pago-modules1/ext/vctr/test/TestCases/TestUtils.cpp:62: Failure:
  REQUIRE_THAT(vWithin, vctr::Equals (reference).withEpsilon (epsilon))
with expansion:
  vctr::Array<float, 10> with size 10 and content [ 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1, 1.1 ] Equals: vctr::Array<float, 10> with size 10 and content [ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 ]
```

pmglr: This might be because vctr::Equals uses Catch::Approx internally. Maybe this should be updated to WithinAbs and WithinRel

PR #73, opened 2023-03-01
Linspace, forEach: make sure they can be constexpr-evaluated

pmglr:


markus-sonible (approved):
looks fine to me!

include/vctr/Containers/VctrBase.h
```diff
@@ -20,6 +20,11 @@
   ==============================================================================
 */
 
+#if VCTR_MSVC
+__pragma(warning(push))
+    __pragma(warning(disable:4324))
```
JanosGit: Funky indentation..?

include/vctr/Containers/VctrBase.h
```diff
@@ -303,7 +308,7 @@ class VctrBase : private Config,
 
         const auto increment = (stop - start) / double (num - size_t (includeEnd));
 
-        [[maybe_unused]] const auto isIntegerIncrement = increment - std::floor (increment) == ElementType (0);
+        [[maybe_unused]] const auto isIntegerIncrement = increment - gcem::floor (increment) == ElementType (0);
```
JanosGit, line 315: GCEM is only an optional dependency, so every code using it should provide a fallback solution for cases when GCEM is not available

JanosGit (changes requested):
Good additions, but please make sure that this also works without GCEM

include/vctr/Containers/VctrBase.h
pmglr: Oops

pmglr: The indentation is actually clang-format conform. However not the missing spaces

JanosGit: I'd rather wrap that in clang-format guards and align it manually, don't see any good reason why this should look like this, I rather consider this a clang-format fault

JanosGit (approved):

PR #74, opened 2023-03-01
Vctr 30 implement assignment operators

### LINKED ISSUE #14: Implement assignment operators

JanosGit:
I tried to implement the assignment operators in a way that a move is preferred in case the source is moved or passed by value, while defaulting back to a plain element copy otherwise

### DIFF
--- a/include/vctr/Containers/Span.h
+++ b/include/vctr/Containers/Span.h
@@ -103,11 +103,30 @@ public:
         VCTR_ASSERT (extent == containerToView.size() || extent == std::dynamic_extent);
     }
 
-    /** Copies the data of the source container to the memory this span refers to. */
+    //==============================================================================
+    // Operators
+    //==============================================================================
+    /** Copies or moves the data of the source container to this Span. You have to ensure that the source size matches. */
     template <has::sizeAndDataWithElementType<ElementType> Container>
-    constexpr Span& operator= (const Container& containerToCopyDataFrom)
+    constexpr Span& operator= (Container&& containerToCopyDataFrom)
     {
-        Vctr::copyFrom (containerToCopyDataFrom.data(), containerToCopyDataFrom.size());
+        if constexpr (Vctr::template shouldMoveFromOtherContainer<Container>)
+        {
+            VCTR_ASSERT (containerToCopyDataFrom.size() == Vctr::size());
+            std::copy (std::make_move_iterator (containerToCopyDataFrom.begin()), std::make_move_iterator (containerToCopyDataFrom.end()), Vctr::storage.begin());
+        }
+        else
+        {
+            Vctr::copyFrom (containerToCopyDataFrom.data(), containerToCopyDataFrom.size());
+        }
+
+        return *this;
+    }
+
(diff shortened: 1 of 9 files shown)


include/vctr/Containers/Span.h
```diff
@@ -103,11 +103,34 @@ class Span : public VctrBase<ElementType, std::span<ElementType, extent>, extent
...
+    /** Copies or moves the data of the source container to this Span. You have to ensure that the source size matches. */
     template <has::sizeAndDataWithElementType<ElementType> Container>
-    constexpr Span& operator= (const Container& containerToCopyDataFrom)
+    constexpr Span& operator= (Container&& containerToCopyDataFrom)
+    {
+        if constexpr (has::begin<Container> &&
+                      has::end<Container> &&
+                      (! is::view<Container>) &&
+                      (! std::is_reference_v<Container>) &&
+                      (! std::is_trivially_copyable_v<ElementType>))
```
p4th4x0r: worth to combine those concepts, no? :)

include/vctr/Containers/Vector.h
```diff
@@ -200,11 +200,40 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+    /** Copies or moves the data of the source container to this Vector. This Vector will be resized to fit if needed. */
     template <has::sizeAndDataWithElementType<ElementType> Container>
-    constexpr Vector& operator= (const Container& containerToCopyDataFrom)
+    constexpr Vector& operator= (Container&& containerToCopyDataFrom)
     {
-        Vctr::copyFrom (containerToCopyDataFrom.data(), containerToCopyDataFrom.size());
+        if constexpr (has::begin<Container> &&
+                      has::end<Container> &&
+                      (! is::view<Container>) &&
+                      (! std::is_reference_v<Container>))
```
p4th4x0r, line 210: again here the same concept. we should combine those :)

include/vctr/Containers/Vector.h
```diff
@@ -200,11 +200,40 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
     //==============================================================================
     // Operators
     //==============================================================================
-    /** Copies the data of the source container to this Vector. This Vector will be resized to fit if needed. */
+    /** Copies or moves the data of the source container to this Vector. This Vector will be resized to fit if needed. */
     template <has::sizeAndDataWithElementType<ElementType> Container>
-    constexpr Vector& operator= (const Container& containerToCopyDataFrom)
+    constexpr Vector& operator= (Container&& containerToCopyDataFrom)
```
p4th4x0r, line 205: so assigning an existing container still works? yep, obviously :)

p4th4x0r (approved):

include/vctr/Containers/Span.h
JanosGit: I combined it in a protected bool variable template now

include/vctr/Containers/Vector.h
JanosGit: Yes, this only makes it possible to additionally move values if we move assign some container

PR #79, opened 2023-03-30
Vctr 54 implement pow

### LINKED ISSUE #28: Implement pow

JanosGit:
I added some macros to avoid repeating boilerplate code for binary expressions. I'd like to do some further refactoring for the existing `VCTR_COMMON_UNARY_EXPRESSION_MEMBERS` macro and rework the handling of SIMD registers in `VCTR_COMMON_BINARY_SINGLE_VEC_EXPRESSION_MEMBERS`, but decided to introduce this changes step by step along with the upcoming PRs to not bloat the diff too much ;)

### DIFF
--- a/include/vctr/Expressions/ExpressionTemplate.h
+++ b/include/vctr/Expressions/ExpressionTemplate.h
@@ -150,7 +150,7 @@ void tryApplyingRuntimeArgsToSrc (const RuntimeArgs& args, Src& src)
 
 } // namespace vctr
 
-/** A helper macro to avoid repetitive boilerplate code when implementing an expression template class.
+/** A helper macro to avoid repetitive boilerplate code when implementing a unary expression template class.
 
     It expects that the expression class has a suitable value_type defined before this, that the second
     class template type name is SrcType and that there is a private member SrcType src.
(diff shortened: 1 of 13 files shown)


JanosGit:
Not really sure what these linux build errors are about – seems to be some internals of `std::pow` being unable to resolve a correct `std::abs` overload that's used inside. However, I think all the test types used here should be valid for `std::pow`. Any idea?

include/vctr/Expressions/ExpressionTemplate.h
```diff
@@ -177,4 +177,85 @@ void iterateOverRuntimeArgChain (const RuntimeArgs& rtArgs)
...
+    using Expression = ExpressionTypes<value_type, SrcAType, SrcBType>;                                                        \
+    using SrcAStorageInfoType = std::invoke_result_t<decltype (&std::remove_cvref_t<SrcAType>::getStorageInfo), SrcAType>;     \
+    using SrcBStorageInfoType = std::invoke_result_t<decltype (&std::remove_cvref_t<SrcBType>::getStorageInfo), SrcBType>;     \
+                                                                                                                               \
+    SrcAType srcAName;                                                                                                         \
+    SrcBType srcBName;                                                                                                         \
+    const CombinedStorageInfo<std::remove_cvref_t<SrcAStorageInfoType>, std::remove_cvref_t<SrcBStorageInfoType>> storageInfo; \
+                                                                                                                               \
+public:                                                                                                                        \
```
TheWaveWarden, line 202: might this cause trouble as we're setting public scope from the macro, which might not be immediately clear when using it?

But then there's no way around it, so probably fine 🤔

TheWaveWarden (approved):

include/vctr/Expressions/ExpressionTemplate.h
JanosGit: I think for this specific use-case this is totally fine. We could of course just make everything public, but I'm not sure if I prefer that. Anyway, I'm going to document that behaviour a bit better.

PR #80, opened 2023-04-03
Vctr 31 implement get underlying vector move underlying vector and implicit conversion operators

### LINKED ISSUE #15: Implement getUnderlyingVector, moveUnderlyingVector and implicit conversion operators

JanosGit:


include/vctr/Containers/VctrBase.h
```diff
@@ -1091,6 +1091,9 @@ class VctrBase : private Config,
     template <class T>
     VCTR_FORCEDINLINE constexpr static T* assumeAlignedToMaxSIMDRegisterSize (T* ptr)
     {
+        if (std::is_constant_evaluated())
+            return ptr;
```
pmglr, line 1095: Can confirm this fixes my issue!

include/vctr/Containers/VctrBase.h
```diff
@@ -1091,6 +1091,9 @@ class VctrBase : private Config,
...
+
         #if __cpp_lib_assume_aligned
```
pmglr: Not new code but this is a rather unusual indentation for a preprocessor directive, no?

include/vctr/Containers/VctrBase.h
pmglr: Why do you omit the `std::assume_aligned` in this case?

include/vctr/Containers/Vector.h
```diff
@@ -257,6 +257,32 @@ class Vector : public VctrBase<ElementType, detail::StdVectorType<ElementType, A
...
+
+       The returned type is std::vector, possibly with a custom allocator type.
+     */
+    constexpr StdVectorType&& moveUnderlyingVector() && { return std::move (Vctr::storage); }
+
+    /** Shortcut to getUnderlyingVector. */
+    constexpr operator StdVectorType&() { return Vctr::storage; }
+
+    /** Shortcut to getUnderlyingVector. */
+    constexpr operator const StdVectorType&() const { return Vctr::storage; }
```
pmglr, line 284: Maybe here you could mention 'conversion operator' as a keyword in the doc for those who are searching the code to finde how it's defined

test/TestCases/VectorMemberFunctions.cpp
pmglr, line 1: Maybe you could also add a test case for span that simply makes sure the following compiles with the changes in `assumeAlignedToMaxSIMDRegisterSize`

```
static constexpr vctr::Array<int, 1> a {0};
static constexpr auto b = a.subSpan<0>();
```

include/vctr/Containers/VctrBase.h
JanosGit: `std::assume_aligned` gives the compiler a hint that it can treat the memory as aligned in order to squeeze out some runtime performance optimisations. However, this function can cause undefined behaviour as the compiler cannot know that the pointed to memory is really aligned, it just trusts us here. If the pointer should be unaligned at runtime, all bets are off and crashes might occur.

In a constant evaluated context such operations that could cause undefined behaviour are forbidden – this is why this addition broke constexpr-compatibility. Furthermore we don't really care about runtime performance optimisation for functions that are evaluated at compile time

include/vctr/Containers/VctrBase.h
JanosGit: Indeed! I'll fix that

include/vctr/Containers/Vector.h
JanosGit: Good point, I'll add that

test/TestCases/VectorMemberFunctions.cpp
JanosGit: I think generally it's a good idea to add a bunch of test cases based on `static_assert` that check all constexpr functions along with the runtime functions. But you are right, I'll add at least some checks for the subSpan

include/vctr/Containers/VctrBase.h
pmglr: I see, thanks for explaining!

pmglr (approved):

PR #81, opened 2023-04-04
Vctr 58 implement cube sqrt

### LINKED ISSUE #8: Implement cube, sqrt

JanosGit:
I did some refactoring of the `VCTR_COMMON_UNARY_EXPRESSION_MEMBERS` macro along the way to align it to the other macros and updated the readme accordingly


JanosGit:
@iromur would you mind testing the unit test artefact on your ARM Mac locally?

include/vctr/Expressions/ExpressionTemplate.h
```diff
@@ -155,28 +155,36 @@ void tryApplyingRuntimeArgsToSrc (const RuntimeArgs& args, Src& src)
...
-template <size_t i, class RuntimeArgs>                                                     \
-void iterateOverRuntimeArgChain (const RuntimeArgs& rtArgs)                                \
-{                                                                                          \
-    tryApplyingRuntimeArgsToThisExpression<i> (rtArgs, *this);                             \
-    tryApplyingRuntimeArgsToSrc<i + 1> (rtArgs, src);                                      \
-}
+#define VCTR_COMMON_UNARY_EXPRESSION_MEMBERS(ExpressionName, srcName)                          \
+private:                                                                                       \
+    using Expression = ExpressionTypes<value_type, SrcType>;                                   \
+    SrcType srcName;                                                                           \
```
pmglr: What is this member used fore?

JanosGit: I'd propose you to have a look at the expression readme for a better overview: https://github.com/sonible/VCTR/blob/f0175800f9defff312c1bc104215462a78746b79/include/vctr/Expressions/Readme.md 

Short answer: This holds the source of this expression, it could either be a reference to the source container or a chained source expression

pmglr: Thank you, I'll check it out!
What I meant though is that it seems you introduced a new member that is not referred to anywhere else in the diff

JanosGit: This is the replacement for all those explicit members that have now been removed in all expressions: 
image

pmglr: But you haven't changed its name to `srcName` anywhere. Or is this merely a member that exists but is not needed anywhere atm?

include/vctr/Expressions/Core/Mean.h
```diff
@@ -228,18 +228,18 @@ namespace vctr
...
+constexpr inline ExpressionChainBuilder<Expressions::MeanSquare> meanSquare;
 
 /** Computes the square root of the mean value of the squared source values.
 
     @ingroup Expressions
  */
-constexpr ExpressionChainBuilder<Expressions::RootMeanSquare> rms;
+constexpr inline ExpressionChainBuilder<Expressions::RootMeanSquare> rms;
 
 } // namespace vctr
```
iromur: 🙀 No newline!

include/vctr/Expressions/ExpressionTemplate.h
pmglr: Also... wouldn't one expect a variable `srcName` to contain the name of a source?

include/vctr/Expressions/Core/Cube.h
```diff
@@ -0,0 +1,82 @@
...
+        return Expression::Neon::mul (x, y);
+    }
+};
+
+} // namespace vctr::Expressions
+
+namespace vctr
+{
+
+/** Raises the source values to the power of three.
```
iromur, line 76: I want specific raise-to-the-power expressions for at least all prime numbers less than 100! 🤖

include/vctr/Expressions/ExpressionTemplate.h
JanosGit: `srcName` is the second macro argument and it has been set to `src` in all the occurences:

image

This just aligns that macro to the ones for operations with multiple sources where more meaningful names like `base` and `exp` for the two sources rather than `srcA` and `srcB` have been chosen. Not sure if we would use other names in the context of unary expressions so much, but I think it just makes the code more unified

include/vctr/Expressions/Core/Square.h
```diff
@@ -40,13 +40,20 @@ class Square : ExpressionTemplateBase
...
     {
         const auto* s = src.evalNextVectorOpInExpressionChain (dst);
         Expression::Accelerate::mul (s, s, dst, size());
         return dst;
     }
 
+    VCTR_FORCEDINLINE const value_type* evalNextVectorOpInExpressionChain (value_type* dst) const
+    requires is::suitableForAccelerateRealFloatVectorOp<SrcType, value_type, detail::dontPreferIfIppAndAccelerateAreAvailable>
+    {
+        Expression::Accelerate::square (src.evalNextVectorOpInExpressionChain (dst), dst, size());
```
iromur, line 53: Ha! So they have a special FPU instruction for squaring floats? 🔲

include/vctr/Expressions/Core/Sqrt.h
```diff
@@ -0,0 +1,85 @@
...
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (Sqrt, src)
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        VCTR_ASSERT (inputValueValid (src[i]));
+        return std::sqrt (src[i]);
+    }
+
+    //==============================================================================
```
pmglr, line 41: Style: I prefer that, if you use separators like this at all, to also put them after the access specifiers. Not sure though if we are generally not following this style in VCTR

include/vctr/Expressions/ExpressionTemplate.h
```diff
@@ -155,28 +155,36 @@ void tryApplyingRuntimeArgsToSrc (const RuntimeArgs& args, Src& src)
...
+public:                                                                                        \
+                                                                                               \
+    template <class Src>                                                                       \
+    constexpr ExpressionName (Src&& s) : src (std::forward<Src> (s)) {}                        \
```
iromur: Shouldn't all previous occurrences of `src` read `srcName` now (also below), otherwise it'd only work if the template argument is `src` by chance?

include/vctr/Expressions/Readme.md
```diff
@@ -9,20 +9,37 @@ template <size_t extent, class SrcType>
...
+private:
+    // Will either be a reference to a source container, a chained source expression or in rare cases a container
+    // owned by this expression
+    SrcType src;
+    
+public:
+    
+    // A Constructor that forwards any kind of source
+    template <class Src>
+    constexpr ExpressionName (Src&& s) : src (std::forward<Src> (s)) {}
```
iromur: Is there trailing whitespace? 🙀

include/vctr/Expressions/Core/Sqrt.h
JanosGit: I generally agree that this is a good thing. I primarily use them here to separate sections like 
```
Default implementation

Vector Op based implementation 

SIMD Register based implementation 
```


However I see that even here multiple patterns have evolved where in some cases there are also headings like
```
Default implementation
// Platform Vector Operation Implementation
Vector Op based implementation 

// AVX Implementation
AVX based SIMD implementation 

// SSE Implementation
SSE based SIMD implementation 

// Neon Implementation
Neon based SIMD implementation 
```

--> To me it seems that some cleanup and style alignment should be done anyway. I'd like to discuss our preferred style in the team and once we agreed on one refactor all the code and write a little style guide in a separate task

include/vctr/Expressions/Core/Sqrt.h
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+private:
+    /** If the value type is real, this checks if the value is greater or equal zero */
+    static constexpr bool inputValueValid (value_type in)
+    {
+        if constexpr (is::realNumber<value_type>)
+            return in >= value_type (0);
+
+        return true;
```
pmglr, line 70: The input value is always valid if `value_type` is not a real number?

JanosGit: Yes, I think for complex numbers (which is basically all non-real numbers) the input is not restricted to any value range. Am I wrong with that assumption?

include/vctr/TypeTraitsAndConcepts/ContainerAndExpressionConcepts.h
```diff
@@ -276,9 +276,10 @@ concept suitableForIppRealOrComplexFloatVectorReductionOp = detail::isPreferredV
 //==============================================================================
 /** Constrains two source types to be suitable for a an aliasing-free binary vector operation using platform vector ops. */
 template <class SrcA, class SrcB, class DstType>
-concept suitableForBinaryVectorOp = (expressionWithEvalVectorOp < SrcA, DstType > && anyVctr < SrcB >) ||
+concept suitableForBinaryVectorOp = ((expressionWithEvalVectorOp < SrcA, DstType > && anyVctr < SrcB >) ||
                                     (anyVctr<SrcA> && expressionWithEvalVectorOp < SrcB, DstType >) ||
-                                    (anyVctr<SrcA> && anyVctr<SrcB>);
+                                    (anyVctr<SrcA> && anyVctr<SrcB>)) &&
```
iromur: I think the previous two lines would need an indentation by 1 to the right?

include/vctr/Expressions/Readme.md
```diff
@@ -9,20 +9,37 @@ template <size_t extent, class SrcType>
...
+    // destination type
+    using value_type = vctr::ValueType<SrcType>;
     
-    using Expression = ExpressionTypes<value_type, SrcType>;
-
-    constexpr MyExpression (SrcType s) : src (std::move (s)) {}
+private:
+    // Will either be a reference to a source container, a chained source expression or in rare cases a container
+    // owned by this expression
+    SrcType src;
```
pmglr, line 20: You'll also need to change the name here

test/TestCases/Expressions/SquareCube.cpp
```diff
@@ -22,14 +22,41 @@
...
+template <class T>
+T cube (T x) { return x * x * x; }
+
+TEMPLATE_PRODUCT_TEST_CASE ("Square Root", "[sqrt]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double, std::complex<float>, std::complex<double>) )
+{
+    VCTR_TEST_DEFINES_IN_RANGE (0, 10000, 10)
+
+    const vctr::Vector s = filter << vctr::sqrt << srcA;
+
+    REQUIRE_THAT (s, vctr::EqualsTransformedBy<sqrt> (srcA).withEpsilon (0.000001));
```
iromur, line 43: Should the detection of an `sqrt (negativeFloat)` attempt be also tested?

pmglr, line 43: Which is the `sqrt` you're testing against here?

include/vctr/Expressions/Core/Sqrt.h
pmglr: I didn't expect complex numbers to be supported at all. If yes, maybe lets add a test case
Edit: test exists, all good!

include/vctr/Expressions/ExpressionTemplate.h
pmglr: Thx for explaining, I was confused for a moment. Makes sense now.

pmglr (approved):

test/TestCases/Expressions/SquareCube.cpp
JanosGit: Since that would only throw an assertion, this would not be that meaningful imo. In general, I could think of a pattern where we inject an exception for the VCTR_ASSERT macro and test if that has been hit in unit tests. Let's add that to the backlog

include/vctr/TypeTraitsAndConcepts/ContainerAndExpressionConcepts.h
JanosGit: that's right indeed

include/vctr/Expressions/Readme.md
JanosGit: No idea what that is 🤨 I'll investigate

include/vctr/Expressions/Core/Cube.h
JanosGit: Feel free to raise a PR for that 😄

include/vctr/Expressions/Core/Mean.h
JanosGit: 😱

include/vctr/Expressions/Core/Square.h
JanosGit: Actually don't know about all the magic they have hidden in their CPUs. Might map to something super special or just wrap a multiplications, we don't know 🤷‍♂️

include/vctr/Expressions/ExpressionTemplate.h
JanosGit: This is completely right. Thanks for spotting that

PR #84, opened 2023-04-05
Vctr xx minor fixes

iromur:


JanosGit (approved):
Thanks for spotting these errors!

Seems that the RMS test case could need some more tolerance, maybe you want to increase it prior to merging?

PR #85, opened 2023-04-06
Vctr 44 implement transformation expressions

### LINKED ISSUE #20: Implement transformation expressions

JanosGit:
Okay, the `transformedBy` expression uses quite a bit of trickery to forward the function to the expression. I hope you can follow the changes, otherwise feel free to ask me for some more explanation ;)


test/TestCases/Expressions/Transformation.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+TEMPLATE_PRODUCT_TEST_CASE ("TransformedBy", "[transformation]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double, int32_t, int64_t))
+{
+    VCTR_TEST_DEFINES (10)
+
+    int counter = 0;
+
+    vctr::Vector convertedToString          = filter << vctr::transformedBy ([] (auto n) { return std::to_string (n); }) << (srcA + ElementType (1));
+    vctr::Vector convertedToStringWithCount = filter << vctr::transformedBy ([&] (auto n) { ++counter; return std::to_string (n); }) << (srcA + ElementType (1));
```
JanosGit, line 55: This tests for both, a trivially copyable lambda and one with capture. Maybe I should add a comment on that to the test code...

include/vctr/Expressions/ExpressionChainBuilder.h
```diff
@@ -253,4 +253,16 @@ auto makeExpressionChainBuilderWithRuntimeArgs (RuntimeArgs... runtimeArgs)
...
+/** Helper function to build factory functions for expressions that rely on runtime argument values.
+
+    This function is just like makeExpressionChainBuilderWithRuntimeArgs except for allowing to pass one
+    further template type argument to the ExpressionType.
+ */
+template <template <size_t, class...> class ExpressionType, class ExpressionTemplateArg, class... RuntimeArgs>
+requires (sizeof... (RuntimeArgs) > 0)
+auto makeTemplateExpressionChainBuilderWithRuntimeArgs (RuntimeArgs... runtimeArgs)
+{
+    return ExpressionChainBuilderWithRuntimeArgs<ExpressionType, detail::RuntimeArgChain<std::tuple<RuntimeArgs...>>, ExpressionTemplateArg> (detail::RuntimeArgChain<std::tuple<RuntimeArgs...>> (std::forward<RuntimeArgs> (std::move (runtimeArgs))...));
```
iromur: What is the semantics of `std::forward (std::move ())`? Wouldn't `std::forward` be sufficient?

include/vctr/Expressions/Generic/TransformedBy.h
```diff
@@ -0,0 +1,74 @@
...
+    might or might not allocate heap memory. For all trivially copyable types such as simple function pointers or
+    lambdas without any capture, you can expect this expression to not allocate any heap memory. For other types of
+    functions you should expect a copy involving heap memory allocation.
+
+    @ingroup Expressions
+ */
+template <class Fn>
+constexpr auto transformedBy (Fn&& fn)
+{
+    return makeTemplateExpressionChainBuilderWithRuntimeArgs<Expressions::TransformedBy, std::remove_cvref_t<Fn>> (std::forward<Fn>(fn));
```
iromur: Style: blank between `<Fn>` and `(fn)`?

test/TestCases/Expressions/Transformation.cpp
iromur: Yes, that's probably a good idea.
We cannot test for allocating/non-allocating behaviour here, can we?

test/include/vctr_test_utils/Matchers/EqualsTransformedBy.h
```diff
@@ -53,6 +53,13 @@ auto EqualsTransformedBy (T scalar, const ReferenceVec& vec);
 
 namespace detail
 {
+
+template <class T>
+double defaultEpsilon = 0.0;
```
iromur, line 58: I don't really get the idea of that template here… 🤔

include/vctr/Expressions/Generic/TransformedByStaticCast.h
```diff
@@ -0,0 +1,67 @@
...
+        return static_cast<DstValueType> (src[i]);
+    }
+
+    VCTR_FORCEDINLINE const DstValueType* evalNextVectorOpInExpressionChain (DstValueType* dst) const
+    requires (sizeof (DstValueType) == sizeof (ValueType<SrcType>) || is::anyVctr<SrcType>) &&
+             has::evalNextVectorOpInExpressionChain<SrcType, ValueType<SrcType>> &&
+             std::is_trivially_destructible_v<ValueType<SrcType>> &&
+             std::is_trivially_constructible_v<DstValueType>
+    {
+        const auto* s = src.evalNextVectorOpInExpressionChain (static_cast<ValueType<SrcType>*> (static_cast<void*> (dst)));
```
iromur, line 48: This is to silence compiler complaints about unrelated types? Probably worth a comment, but not sure whether I get the full picture…

iromur (approved):
Not sure whether I get the full picture of the trickery but appears reasonable.

include/vctr/Expressions/Generic/TransformedByStaticCast.h
JanosGit: This wouldn't even compile otherwise. 

Generally a chain of `evalNextVectorOpInExpressionChain` calls work by using the destination memory as a buffer for intermediate results when evaluating the chain of expressions. The very first expression reads from the source memory and writes to the destination memory. All following operations take the destination memory as both, source and destination. 

Now let's take this expression as an example
```c++
vctr::Vector<float> a = { 1.0f, 2.0f, 3.0f };
vctr::Vector<int> b = vctr::transformedByStaticCastTo<int> << (a - 1.0f);
```

If evaluated with vector operations this expression would first evaluate `a - 1.0f` using the memory of `b` as the destination for the intermediate result. After that it would evaluate `vctr::transformedByStaticCastTo<int>` using `b` as both, the source and destination memory for the result. However in the specific case of a cast the `dst` pointer has to change its type in order to act as a suitable destination pointer for the float expression. Technically, `int` and `float` are not convertible on a bit level, but we know that using the destination buffer of a different type  works as long as we know that the size of both types is the same. The only way to get the compiler to perform that conversion is by casting it to `void*` first.

test/TestCases/Expressions/Transformation.cpp
JanosGit: Nope, that's not really possible

test/include/vctr_test_utils/Matchers/EqualsTransformedBy.h
JanosGit: What we really want is the specialisation below for numeric types. The default case is just a fallback to make tests with non numeric types compile, since `defaultEpsilon` is used as default argument to the `withEpsilon` member functions.

include/vctr/Expressions/ExpressionChainBuilder.h
JanosGit: Uhm, good spot. Actually the `std::forward` is completely unnecessary here since we don't want any perfect forwarding in that case. This is simply a stupid copy & paste error ;)

include/vctr/Expressions/Generic/TransformedBy.h
JanosGit: Yes!

PR #87, opened 2023-04-11
Vctr 49 implement sort lower bound upper bound is sorted

### LINKED ISSUE #25: Implement sort, lowerBound, upperBound, isSorted

JanosGit:


include/vctr/Containers/VctrBase.h
```diff
@@ -600,6 +600,30 @@ class VctrBase : private Config,
         return it == rend() ? std::nullopt : std::optional<size_t> (std::distance (it, rend()) - 1);
     }
 
+    /** Returns a copy of the first element value which is greater or equal to bound or std::nullopt if no such value is found.
+
+        This function requires that the elements are sorted.
+     */
+    std::optional<value_type> firstValueGreaterThanOrEqualTo (const value_type& bound) const
```
JanosGit: I decided to go for some more descriptive names than `lowerBound` because I always had to really study cppreference for a least one minute to remember what the this function was about and how it is different to upper bound

p4th4x0r: same here. every time using the function i had to recover my knowledge

include/vctr/Containers/VctrBase.h
```diff
@@ -600,6 +600,30 @@ class VctrBase : private Config,
         return it == rend() ? std::nullopt : std::optional<size_t> (std::distance (it, rend()) - 1);
     }
 
+    /** Returns a copy of the first element value which is greater or equal to bound or std::nullopt if no such value is found.
```
p4th4x0r: hm, tbh it does not return a copy of the first element but an optional of the copy of the first element ;)

include/vctr/Containers/VctrBase.h
p4th4x0r: so if you change the name of the function from lower bound to the new one, the parameter name "bound" might be misleading.

include/vctr/Containers/VctrBase.h
```diff
@@ -600,6 +600,30 @@ class VctrBase : private Config,
...
+        VCTR_ASSERT (elementsAreSorted());
+        auto it = std::lower_bound (begin(), end(), bound);
+        return it == end() ? std::nullopt : std::optional<value_type> (*it);
+    }
+
+    /** Returns a copy of the first element value which is greater than bound or std::nullopt if no such value is found.
+
+        This function requires that the elements are sorted.
+     */
+    std::optional<value_type> firstValueGreaterThan (const value_type& bound) const
```
p4th4x0r: also here parameter name "bound" might be misleading.

include/vctr/Containers/VctrBase.h
```diff
@@ -672,6 +696,30 @@ class VctrBase : private Config,
             clear (data(), n);
     }
 
+    /** Sorts all elements in this vector according to the sortingPredicate function. */
```
p4th4x0r: and the sortingPredicate function is what here? :D

p4th4x0r (changes requested):

include/vctr/Containers/VctrBase.h
JanosGit: Agreed this could be more precise

include/vctr/Containers/VctrBase.h
JanosGit: Mhm, what parameter name would you suggest?

include/vctr/Containers/VctrBase.h
JanosGit: Ah copy & paste leftover, will fix that

include/vctr/Containers/VctrBase.h
p4th4x0r: maybe "valueToFind" or just "value" but "bound" means nothing here :)

test/TestCases/VctrBaseMemberFunctions.cpp
```diff
@@ -368,6 +368,42 @@ TEST_CASE ("shift", "[VctrBaseMemberFunctions]")
...
+TEST_CASE ("sort", "[VctrBaseMemberFunctions]")
+{
+    vctr::Array numericalValues { -3, 10, 78, -24, 0 };
+    REQUIRE_FALSE (numericalValues.elementsAreSorted());
+    numericalValues.sort();
+    REQUIRE_THAT (numericalValues, vctr::Equals ({ -24, -3, 0, 10, 78}));
+    REQUIRE (numericalValues.elementsAreSorted());
+
+    std::string characters = "Ac/Dc";
+    vctr::Span (characters).sort ([] (char lhs, char rhs) { return std::tolower (lhs) < std::tolower (rhs); });
```
JanosGit, line 380: Side note: I was pretty happy to realise that I could simply wrap a `std::string` in a `vctr::Span` like that and then treat it as a char array ☺️

p4th4x0r (approved):

PR #88, opened 2023-04-12
Vctr 56 implement map normalize

### LINKED ISSUE #30: Implement map, normalize

JanosGit:
Note: This includes all the commits from https://github.com/sonible/VCTR/pull/87 since the unit tests use the `sort` function implemented there. Should be rebased on develop as soon as the other PR has been merged

### DIFF
--- a/include/vctr/Expressions/BasicMath/NormalizeSum.h
+++ b/include/vctr/Expressions/BasicMath/NormalizeSum.h
@@ -1,0 +1,88 @@
+/*
+  ==============================================================================
+    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+
+    Copyright 2023 by sonible GmbH.
+
+    This file is part of VCTR - Versatile Container Templates Reconceptualized.
+
+    VCTR is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License version 3
+    only, as published by the Free Software Foundation.
+
+    VCTR is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License version 3 for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
+template <size_t extent, class SrcType>
+requires is::floatNumber<ValueType<SrcType>>
+class NormalizeSum : ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (NormalizeSum, src)
+
+    void applyRuntimeArgs()
+    {
+        srcSum = sum << src;
+    }
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        VCTR_ASSERT (srcSum != value_type (0));
+        return src[i] / srcSum;
+    }
+
+    //==============================================================================
+    // Platform Vector Operation Implementation
+    VCTR_FORCEDINLINE const value_type* evalNextVectorOpInExpressionChain (value_type* dst) const
+    requires is::suitableForAccelerateRealFloatVectorOp<SrcType, value_type, detail::dontPreferIfIppAndAccelerateAreAvailable>
+    {
+        VCTR_ASSERT (srcSum != value_type (0));
+        Expression::Accelerate::div (src.evalNextVectorOpInExpressionChain (dst), srcSum, dst, size());
+        return dst;
+    }
+
+    VCTR_FORCEDINLINE const value_type* evalNextVectorOpInExpressionChain (value_type* dst) const
+    requires is::suitableForIppRealFloatVectorOp<SrcType, value_type, detail::preferIfIppAndAccelerateAreAvailable>
+    {
+        VCTR_ASSERT (srcSum != value_type (0));
+        Expression::IPP::div (src.evalNextVectorOpInExpressionChain (dst), srcSum, dst, size());
+        return dst;
+    }
+
+private:
+    //==============================================================================
+    value_type srcSum = value_type (0);
+};
+
+}
+
+namespace vctr
+{
+
+/** Computes the sum of the source elements and divides all source elements by that sum.
+
+    Note: If the source is an expression itself, that source expression will possibly be evaluated twice
+    since we need to evaluate it once to compute the sum and then again to compute the normalized expression
+    result. For performance critical applications, you should consider buffering a source expression to an
+    intermediate container.
+
+    @attention The required sum is computed once when the source is assigned to the expression. In case
+    the source values are changed after that time point, evaluating the expression will use an outdated
+    sum. Therefore, we might consider deprecating and eventually removing this expression in a future release.
+
+    @ingroup Expressions
+ */
+constexpr inline ExpressionChainBuilder<expressions::NormalizeSum> normalizeSum;
+}
--- a/include/vctr/Expressions/ExpressionTemplate.h
+++ b/include/vctr/Expressions/ExpressionTemplate.h
@@ -137,6 +137,12 @@ void tryApplyingRuntimeArgsToThisExpression (const RuntimeArgs& args, Expression
                         e.applyRuntimeArgs (args...);
                     }, args.template get<i>());
     }
+    else if constexpr (has::applyRuntimeArgs<Expression>)
+    {
+        // Some expressions need to initialize runtime values that depend on their source before every
+        // evaluation rather than being injected as external arguments (see e.g. NormalizeSum).
+        e.applyRuntimeArgs();
+    }
 }
 
 template <size_t i, class RuntimeArgs, is::anyVctrOrExpression Src>
--- a/include/vctr/Expressions/Generic/Map.h
+++ b/include/vctr/Expressions/Generic/Map.h
@@ -1,0 +1,257 @@
+/*
+  ==============================================================================
+    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+
+    Copyright 2023 by sonible GmbH.
+
+    This file is part of VCTR - Versatile Container Templates Reconceptualized.
+
+    VCTR is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Lesser General Public License version 3
+    only, as published by the Free Software Foundation.
+
+    VCTR is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License version 3 for more details.
+
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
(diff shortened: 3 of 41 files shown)


p4th4x0r (approved):

include/vctr/Expressions/BasicMath/NormalizeSum.h
```diff
@@ -0,0 +1,81 @@
...
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
+template <size_t extent, class SrcType>
+requires is::floatNumber<ValueType<SrcType>>
```
p4th4x0r, line 27: so, can't really remember but "floatNumber" is float and double? or also long double etc.

include/vctr/Expressions/BasicMath/NormalizeSum.h
```diff
@@ -0,0 +1,81 @@
...
+class NormalizeSum : ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (NormalizeSum, src)
+
+    void applyRuntimeArgs()
```
p4th4x0r, line 35: at which point is that function executed? it has to be executed before operator[].

include/vctr/Expressions/BasicMath/NormalizeSum.h
```diff
@@ -0,0 +1,81 @@
...
+    {
+        srcSum = sum << src;
+    }
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        return src[i] / srcSum;
```
p4th4x0r, line 43: maybe assert that srcSum != 0?

include/vctr/Expressions/BasicMath/NormalizeSum.h
```diff
@@ -0,0 +1,81 @@
...
+    VCTR_FORCEDINLINE const value_type* evalNextVectorOpInExpressionChain (value_type* dst) const
+    requires is::suitableForIppRealFloatVectorOp<SrcType, value_type, detail::preferIfIppAndAccelerateAreAvailable>
+    {
+        Expression::IPP::div (src.evalNextVectorOpInExpressionChain (dst), srcSum, dst, size());
+        return dst;
+    }
+
+private:
+    //==============================================================================
+    value_type srcSum = value_type (1);
```
p4th4x0r: hmm, i understand why you chose 1 as default value. however, maybe zero would be also a good choice because then you could add that zero assertion i meant before that tells the user something went wrong when calculating the sum :) os in other words, if the srcSum is still 0 when the operator[] is called, something went wrong.

or maybe an optional? so srcSum is nullopt by default. when srcSum is assigned it only becomes non-nullopt when the value is != 0 :)

include/vctr/Expressions/BasicMath/NormalizeSum.h
```diff
@@ -0,0 +1,81 @@
...
+class NormalizeSum : ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (NormalizeSum, src)
+
+    void applyRuntimeArgs()
+    {
+        srcSum = sum << src;
```
p4th4x0r, line 37: at least here, we have to make sure (or assert) that srcSum is != 0

include/vctr/Expressions/ExpressionTemplate.h
```diff
@@ -137,6 +137,12 @@ void tryApplyingRuntimeArgsToThisExpression (const RuntimeArgs& args, Expression
                         e.applyRuntimeArgs (args...);
                     }, args.template get<i>());
     }
+    else if constexpr (has::applyRuntimeArgs<Expression>)
+    {
+        // Some expressions need to initialize runtime values that depend on their source before every
+        // evaluation rather than being injected as external arguments (see e.g. NormalizeSum).
+        e.applyRuntimeArgs();
```
p4th4x0r, line 144: ah okay, here :) did you add that thingy to the docu how to add expressions?

include/vctr/Expressions/Generic/Map.h
```diff
@@ -0,0 +1,260 @@
...
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
+template <size_t extent, class SrcType, is::rangeWithValueType<ValueType<SrcType>> RangeType>
+requires is::floatNumber<ValueType<SrcType>>
+class Map : ExpressionTemplateBase
```
p4th4x0r, line 28: hmm, i'm not 100% convinced of the word "Map" because at the first glance something like a std::map comes into my mind. but within the right namespace it might be okay :)

include/vctr/Expressions/Generic/Map.h
```diff
@@ -0,0 +1,260 @@
...
+
+    You should have received a copy of the GNU Lesser General Public License
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
+template <size_t extent, class SrcType, is::rangeWithValueType<ValueType<SrcType>> RangeType>
```
p4th4x0r, line 26: i do not have in mind that rangeWithValueType checks but note that a juce range does not allow that the end value is lower than the start value ;)

include/vctr/Expressions/Generic/Map.h
```diff
@@ -0,0 +1,260 @@
...
+        srcDstLenRatio = newDstRange.getLength() / newSrcRange.getLength();
+        dstRangeStart = newDstRange.getStart();
+    }
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        auto v = src[i] + srcRangeStartNegated;
+        v *= srcDstLenRatio;
+        v += dstRangeStart;
+        return v;
```
p4th4x0r: okay in this case for me
`return srcDstLenRatio * (src[i] + srcRangeStartNegated) + dstRangeStart` 
would be better to read.

include/vctr/Expressions/Generic/Map.h
```diff
@@ -0,0 +1,260 @@
...
+private:
+    //==============================================================================
+    value_type srcRangeStartNegated = 0;
+    value_type srcDstLenRatio = 0;
+    value_type dstRangeStart = 0;
+};
+
+template <size_t extent, class SrcType, is::rangeWithValueType<ValueType<SrcType>> RangeType>
+requires is::floatNumber<ValueType<SrcType>>
+class MapFrom0To1 : ExpressionTemplateBase
```
p4th4x0r, line 82: wouldn't it be much easier to derive from Map where the source range is just zero to one?

include/vctr/Expressions/Generic/Map.h
```diff
@@ -0,0 +1,260 @@
...
+/** Maps all source element values from the range srcValueRangeStart to srcValueRangeEnd to values in [dstValueRangeStart, dstValueRangeEnd].
+
+    The expression does not check if the source values are actually in the expected range.
+
+    @ingroup Expressions
+ */
+template <is::floatNumber T>
+auto map (T srcValueRangeStart, T srcValueRangeEnd, T dstValueRangeStart, T dstValueRangeEnd)
+{
+    return makeTemplateExpressionChainBuilderWithRuntimeArgs<expressions::Map, Range<T>> (Range (srcValueRangeStart, srcValueRangeEnd), Range (dstValueRangeStart, dstValueRangeEnd));
```
p4th4x0r: why not 
`return map (Range (srcValueRangeStart, srcValueRangeEnd), Range (dstValueRangeStart, dstValueRangeEnd));`

p4th4x0r (approved):
Approved besides the small remarks :)

include/vctr/Expressions/BasicMath/NormalizeSum.h
JanosGit: It forwards to https://en.cppreference.com/w/cpp/types/is_floating_point so yes, `float`, `double`, `long double` and hypothetically even fancy new types like `std::float16_t`, `std::float32_t`, `std::float64_t`, `std::float128_t`, or `std::bfloat16_t` that have been introduced with C++23. Or in other words: All non integer real valued numerical types 😉

include/vctr/Expressions/BasicMath/NormalizeSum.h
JanosGit: `applyRuntimeArgs()` is called as soon as a source container is prepended to a (chain of) expression(s). Before, it has been used to pass back runtime values like e.g. the bounds for the `clamp` expression or the range for the `map` expression. I figured out that in this particular case we need the option to compute the sum once before each call to `operator[]` or the other evaluation functions, so I also made it possible that this function can be used without any arguments.

Still you got a point here. There is one particular edge case here which is this example:
```c++
vctr::Vector v { 1.0, 2.0, 3.0 };

auto normalizedSum = vctr::normalizeSum << v;

vctr::Vector res1 = normalizedSum;

v.push_back (4.0);
v[2] = 5.0;

// At this point the normalizedSum expression will still hold the sum of the elements of v from the time point
// where the expression has been created. It will not re-evaluate it here because it does not assume that the
// source values have changed. 
vctr::Vector res2 = normalizedSum;
```

The pattern sketched above where an unevaluated expression is stored to a variable and then is evaluated multiple times while changing values of the source container(s) should just work with all other expressions so far. Although this is a relatively uncommon use-case I think it is actually a flaw of this particular expression that is hard to avoid if we don't want to re-evaluate the sum on each invocation of operator[] which would be a nightmare performance wise.

Tbh. I'm not sure if this expression is more harmful than useful, given that code using it could easilly be rewritten using a straightforward sum and division. We could keep it and add a big note regarding that, we could keep it and declare it deprecated from the start on just to make it easier to rework old code relying on the old `normalizeSum` implementation or we could just not add it and just rework current internal code using `normalizeSum` when rolling out VCTR in our internal code base. What do you think?
I added it

include/vctr/Expressions/BasicMath/NormalizeSum.h
JanosGit: good point

include/vctr/Expressions/BasicMath/NormalizeSum.h
JanosGit: I think 0 is a fine choice since asserting a non-zero value would also catch cases where the computed sum would actually be 0

include/vctr/Expressions/Generic/Map.h
JanosGit: It aligns with the old implementation which was in turn heavily influenced by `juce::jmap`. I'm open for other naming suggestions, but can't really make up a good alternative

include/vctr/Expressions/Generic/Map.h
JanosGit: The `vctr::Range` doesn't allow that as well. Tried to implement it as a minimal lightweight alternative to the `juce::Range`

include/vctr/Expressions/Generic/Map.h
JanosGit: I'm fine with either way

include/vctr/Expressions/Generic/Map.h
JanosGit: I thought about that as well but figured out that we can omit an entire operation in this case which seemed enough reason to go for an extra implementation here on the never ending hunt for performance ;)

include/vctr/Expressions/Generic/Map.h
JanosGit: because I din't think of it ;) Can change that

include/vctr/Expressions/ExpressionTemplate.h
JanosGit: Ah this comment was hidden for some reason. Good call, I'll add some docu

PR #89, opened 2023-04-13
Config: Allow to explicitly disable the member alignment of vctr::Array

JanosGit:


include/vctr/SIMD/SIMDHelpers.h
```diff
@@ -122,10 +122,10 @@ struct StorageInfo<std::array<ElementType, size>>
 {
     constexpr StorageInfo init (const void*, size_t) { return *this; }
 
-    static constexpr size_t memberAlignment = Config::maxSIMDRegisterSize;
+    static constexpr size_t memberAlignment = Config::alignedArray ? Config::maxSIMDRegisterSize : alignof (std::array<ElementType, size>);
 
-    /** This is ensured by the enforced member alignment */
-    static constexpr bool dataIsSIMDAligned = true;
+    /** This might be ensured at build time by an enforced member memberAlignment */
```
psciri: probably add reference to macro here

psciri (commented):
are there any tests in place that verify that the alignment macro can safely be used for both scenarios?

JanosGit:
> are there any tests in place that verify that the alignment macro can safely be used for both scenarios?

I only verified that locally so far. We could add another build to the matrix but I'd rather leave testing of this configuration up to our internal unit tests since there are already so many of them

include/vctr/SIMD/SIMDHelpers.h
JanosGit: Good point

include/vctr/Containers/Span.h
```diff
@@ -106,6 +106,9 @@ class Span : public VctrBase<ElementType, std::span<ElementType, extent>, extent
     //==============================================================================
     // Operators
     //==============================================================================
+    // Deleting the default copy operator forces the usage of the intended operator below
+    Span& operator= (const Span&) = delete;
```
pmglr, line 110: This must be what fixes the problems when assigning one span to another

pmglr: Indeed it fixes the following tests (section 1 and 3 did fail). Can you please add them or equivalent tests?

```
TEST_CASE ("Span assignments")
{
    vctr::Array<int, 1> a (0);
    vctr::Array<int, 1> b (1);

    SECTION ("Attempt 1")
    {
        a.subSpan (0) = b.subSpan (0);
        REQUIRE (a[0] == b[0]); // fails locally
    }

    SECTION ("Attempt 2")
    {
        auto aSpan = vctr::Span (a.data(), a.size());
        auto bSpan = vctr::Span (b.data(), b.size());
        aSpan = bSpan;

        REQUIRE (a[0] == b[0]); // passes locally
    }

    SECTION ("Attempt 3")
    {
        auto makeSpan = [] (int* ptr, size_t size) { return vctr::Span (ptr, size); };

        makeSpan (a.data(), a.size()) = makeSpan (b.data(), b.size());

        REQUIRE (a[0] == b[0]); // fails locally
    }

    SECTION ("Attempt 4")
    {
        auto makeSpan = [] (int* ptr, size_t size) { return vctr::Span (ptr, size); };

        makeSpan (a.data(), a.size())[0] = makeSpan (b.data(), b.size())[0];

        REQUIRE (a[0] == b[0]); // passes locally
    }
}
```

JanosGit: It is. Surprisingly simple 🙈

JanosGit: I added an issue for that, since I'd like to add a few more assignment tests https://github.com/sonible/VCTR/issues/90

psciri (approved):

PR #91, opened 2023-04-19
NumericTypeConcepts: Reworked floating point number concepts

JanosGit:
Rationale: `std::complex` only supports floating point numbers, so having concepts for complex float, complex int and general complex numbers was wrong in the first place. With this change there is only a concept for complex float numbers. Furthermore, the floatNumber concept has been renamed to realFloatNumber and a realOrComplexFloatNumber concept has been added


pmglr (approved):

pmglr:
> Rationale: `std::complex` only supports floating point numbers

You got a point here! 😉

PR #92, opened 2023-04-21
Vctr 59 implement trigonometric functions

### LINKED ISSUE #7: Implement trigonometric functions

JanosGit:


JanosGit:
It was tempting to define a macro for all those expressions since the code is so repetitive, but I decided to stay explicit ;)

include/vctr/Expressions/Trigonometric/Acos.h
```diff
@@ -0,0 +1,90 @@
...
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (Acos, src)
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+#if VCTR_USE_GCEM
+        if constexpr (is::realFloatNumber<value_type>)
+        {
+            if (std::is_constant_evaluated())
```
p4th4x0r, line 40: why is it necessary to request this? i mean even if this is not constant evaluated, we could use gcem, right?

include/vctr/Expressions/Trigonometric/Acosh.h
```diff
@@ -0,0 +1,90 @@
...
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (Acosh, src)
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+#if VCTR_USE_GCEM
+        if constexpr (is::realFloatNumber<value_type>)
+        {
+            if (std::is_constant_evaluated())
+                return gcem::acosh (src[i]);
```
p4th4x0r, line 41: also here?

p4th4x0r, line 41: also here?

include/vctr/Expressions/Trigonometric/Tanh.h
p4th4x0r, line 1: do you have also a atan2 implementation? :)

p4th4x0r (approved):

include/vctr/Expressions/Trigonometric/Acos.h
JanosGit: gcem proved to be slower compared to the std implementation in non constant evaluated cases in the past. Furthermore, in contrast to the std version, it unfortunately does not support complex numbers, so we need the std fallback version in any case. This seemed to be the most straightforward way of chosing the appropriate implementation to me

include/vctr/Expressions/Trigonometric/Tanh.h
JanosGit: > do you have also a atan2 implementation? :)

Ehm not yet. However this is so rarely used that I'd rather add that to the backlog and implement it whenever it should be needed in future. It also falls a bit out of this pattern since it needs two arguments rather than one

pmglr (approved):
Lot's of repetition, but very nice, glad to finally have trigonometric functions!

pmglr:
> It was tempting to define a macro for all those expressions since the code is so repetitive, but I decided to stay explicit ;)

Understandable. We can keep that in mind in case some refactoring will be needed in the future.

PR #95, opened 2023-05-04
ExpressionChainBuilder: declare hasValue constexpr

pmglr:
Reason: consteval did not compile under certain conditions

### DIFF
--- a/include/vctr/Expressions/ExpressionChainBuilder.h
+++ b/include/vctr/Expressions/ExpressionChainBuilder.h
@@ -58,7 +58,7 @@ public:
     }
 
     template <size_t idx>
-    static consteval bool hasValue()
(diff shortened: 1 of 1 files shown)


JanosGit (changes requested):
The class that has been changed is actually named RuntimeArgChain, so I'd propose to reword the commit message to "RuntimeArgChain: declare hasValue constexpr". Otherwise approved

pmglr:
> The class that has been changed is actually named RuntimeArgChain, so I'd propose to reword the commit message to "RuntimeArgChain: declare hasValue constexpr". Otherwise approved

Ah, of course. Thanks. Changed it!

JanosGit (approved):

PR #96, opened 2023-05-08
Sgat 407

JanosGit:


include/vctr_utils/vctr_juce_helpers.h
```diff
@@ -20,6 +20,15 @@
...
 
+#pragma once
+
+#include "../vctr/vctr.h"
+#include <juce_dsp/juce_dsp.h>
+
+/** Helper functions to integrate vctr into dsp code written with JUCE (https://github.com/juce-framework/JUCE).
+
+    It requires that the project links against the JUCE dsp module.
+ */
```
iromur, line 31: Ok, so this will need to be included manually if vctr is going to be used with JUCE, no automagic preprocessor header check? Sounds much cleaner!

iromur (approved):

include/vctr_utils/vctr_juce_helpers.h
JanosGit: Yes excactly. Figured out that the previous approach was rather difficult to integrate, since it relied on the juce dsp module which might not be available at the point where VCTR is included in general

PR #106, opened 2023-06-21
Vctr 98 integrate vctr unit tests in internal tests

### LINKED ISSUE #103: Unknown compiler warning ignored when compiling with gcc
When compiling my project under Ubuntu with gcc, I get a lot of warnings regarding an unknown compiler warning `return-stack-address`. And yes, this warning doesn't seem to exist for gcc: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html :slightly_smiling_face: 
image

JanosGit:


iromur (approved):
Looks good!

include/vctr/TypeTraitsAndConcepts/Traits.h
```diff
@@ -181,9 +181,9 @@ class DataType
 {
 private:
     // This function is only used to deduce a return type, so we can safely ignore warnings about returning a stack address
-    VCTR_START_IGNORE_WARNING_CLANG_GCC (return-stack-address)
+    VCTR_START_IGNORE_WARNING_CLANG (return-stack-address)
```
TheWaveWarden, line 184: Thanks! 👍 😸

test/TestCases/Expressions/Map.cpp
```diff
@@ -25,6 +25,14 @@
 template <int srcMin, int srcMax, int dstMin, int dstMax, vctr::is::realNumber T>
 T map (T val)
 {
+    constexpr auto tMin = std::numeric_limits<T>::lowest();
+    constexpr auto tMax = std::numeric_limits<T>::max();
+
+    static_assert (srcMin >= tMin && srcMin <= tMax);
+    static_assert (srcMax >= tMin && srcMax <= tMax);
+    static_assert (dstMin >= tMin && dstMin <= tMax);
+    static_assert (dstMax >= tMin && dstMax <= tMax);
```
TheWaveWarden, line 34: uhm what does this assert exactly if the limits are lowest and max for that type? NaN?

test/TestCases/VctrBaseMemberFunctions.cpp
```diff
@@ -189,7 +189,7 @@ TEST_CASE ("forEach", "[VctrBaseMemberFunctions]")
     auto strings = loremIpsum;
 
     // In place for each function
-    strings.forEach ([] (std::string& s) { for (auto& c : s) c = std::toupper (c); });
+    strings.forEach ([] (std::string& s) { for (auto& c : s) c = char (std::toupper (c)); });
```
TheWaveWarden, line 192: aaah to upper.. I just googled wondering what a toupper is 😄 strange the standard didn't name this `std::to_upper()`..

TheWaveWarden (approved):
Lookin good! 🙂

test/TestCases/Expressions/Map.cpp
JanosGit: The background here was that in a certain case the type was `int16_t` while the min and max values specified as `int` template arguments were greater than the supported value range for that type

test/TestCases/VctrBaseMemberFunctions.cpp
JanosGit: I think this is a super ancient C function (which also explains the int return type) that just made its way to the C++ `std` namespace. Nothing that I'd consider for usage in production code but here it was suitable to create a compact test case

TheWaveWarden: Makes sense!

test/TestCases/Expressions/Map.cpp
TheWaveWarden: Ah ok I didn't realise these can have differen types. Makes sense then!

PR #107, opened 2023-06-27
Vector: Add constructor that takes an allocator

JanosGit:


JanosGit:
Looking at the test fails it seems like gcc tries to use the new allocator constructor when creating a `Vector` from a `std::list` in the unit cases - don't know why it does so

iromur:
Hmm, cannot say much here… But happy to approve if that makes sense.

JanosGit:
Still don't understand it entirely, but I found a fix for it and also added a unit test for the new constructor along the way

iromur (approved):

PR #108, opened 2023-07-18
Array: Fixed gcc compile errors when assigning a std::vector 

JanosGit:


include/vctr/Containers/Array.h
```diff
@@ -89,7 +89,7 @@ class Array : public VctrBase<ElementType, std::array<ElementType, storageExtent
     {}
 
     /** Creates an Array with extent 1 from an initial value. */
-    template <is::suitableSingleElementInitializer T>
+    template <std::convertible_to<ElementType> T>
```
TheWaveWarden, line 92: Assuming this also fixes the assignment of `std::array` to `vctr::Array` (which was my original issue)? 🙂

TheWaveWarden (approved):

include/vctr/TypeTraitsAndConcepts/ContainerAndExpressionConcepts.h
```diff
@@ -358,12 +358,6 @@ concept inputIteratorToConstructValuesOfType = std::input_iterator<T> && std::co
 template <class T, class ElementType>
 concept contiguousIteratorWithValueTypeSameAs = std::contiguous_iterator<T> && std::same_as<ElementType, std::iter_value_t<T>>;
 
-/** Constrains the type to be suitable for initializing a single element Vctr, that is, it is not of any other type
-    for which special single argument constructors exist, so no ambiguity can arise.
- */
-template <class T>
-concept suitableSingleElementInitializer = ! anyVctrOrExpression<T> && ! triviallyCopyableWithDataAndSize<T> && ! suitableInitializerFunction<T>;
```
JanosGit, line 365: I'm actually not sure why I decided to chose this rather specific blacklist approach back then instead of the more straightforward approach that I have chosen now ;)

include/vctr/Containers/Array.h
JanosGit: Ah I thought it was `std::vector` - but I just added a test case for `std::array` locally and it seems to work as well. I'll push the test additional case

TheWaveWarden: awesome!

Issue #111, opened 2023-07-19
ExpressionChainBuilder breaks runtime arg chain when used with ExpressionType that is a concatenation of expressions itself


JanosGit:
This happened e.g. with `MagToDb` and `DbToMag`. A current quick fix is in place but that's a rather convoluted and non-generic workaround

PR #112, opened 2023-07-19
Expressions: Quick fix for an error where runtime arguments were not applied when a decibel expression was prepended

JanosGit:
Should be fixed with a generic long term solution https://github.com/sonible/VCTR/issues/111

### DIFF
--- a/include/vctr/Expressions/DSP/Decibels.h
+++ b/include/vctr/Expressions/DSP/Decibels.h
@@ -82,7 +82,7 @@ struct dBPower : Constant<10> {};
    @ingroup Expressions
  */
 template <is::constant DecibelConstant, auto minDb = -100>
-constexpr inline ExpressionChainBuilder<expressions::MagToDb, DecibelConstant, Constant<minDb>> magToDb;
+constexpr inline ExpressionChainBuilderWithRuntimeArgs<expressions::MagToDb, detail::RuntimeArgChain<std::tuple<>, std::tuple<>, std::tuple<>>, DecibelConstant, Constant<minDb>> magToDb;
 
 /** Converts the source decibel values into their magnitude representation.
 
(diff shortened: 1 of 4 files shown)


include/vctr/Expressions/DSP/Decibels.h
```diff
@@ -82,7 +82,7 @@ struct dBPower : Constant<10> {};
    @ingroup Expressions
  */
 template <is::constant DecibelConstant, auto minDb = -100>
-constexpr inline ExpressionChainBuilder<expressions::MagToDb, DecibelConstant, Constant<minDb>> magToDb;
+constexpr inline ExpressionChainBuilderWithRuntimeArgs<expressions::MagToDb, detail::RuntimeArgChain<std::tuple<>, std::tuple<>, std::tuple<>>, DecibelConstant, Constant<minDb>> magToDb;
```
JanosGit, line 85: This is the quick fix: If the expression does not need any runtime args, `ExpressionChainBuilder` just creates a dummy `RuntimeArgChain` with a single empty tuple. Since these expressions are compositions actually made up of 3 expressions, manually specifying a `RuntimeArgChain` with three empty tuples fixes it. In the long term, `ExpressionChainBuilder` should be able to figure out if the expression type passed in is a composition of multiple expressions and do that automatically, since this syntax is super confusing

include/vctr/Expressions/ExpressionChainBuilder.h
```diff
@@ -46,10 +46,16 @@ class RuntimeArgChain
...
-    : chain {{ std::tuple<>() }}
+        : chain {{ std::tuple<>() }}
     {}
 
-    static consteval size_t size() { return sizeof... (ArgTuples); }
+    // TODO: This is a workaround for https://github.com/sonible/VCTR/issues/111 and should be replaced by a generic solution
+    constexpr RuntimeArgChain()
+    requires (sizeof... (ArgTuples) == 3)
+        : chain {{ std::tuple<>() }, { std::tuple<>() }, { std::tuple<>() }}
+    {}
```
JanosGit, line 56: This is a quick hardcoded solution for the case above where we need to support 3 empty tuples and a constructor without arguments. Should be replaced by a generic solution in the long run, which could be a bit tricky to get right, so I decided to go for the non-generic solution for now

include/vctr/Expressions/DSP/Decibels.h
psciri: does this break anything? looks like an interface change but cannot really tell from the context..

psciri (approved):

include/vctr/Expressions/DSP/Decibels.h
JanosGit: No this does not break anything. The object relevant to the user is still named `vctr::magToDb` and works correct now if chained with expressions that need the `applyRuntimeArgs` call to work correctly.

PR #114, opened 2023-07-25
Added GH-Pages deploy step to Documentation.yml

psciri:


.github/workflows/Documentation.yml
```diff
@@ -37,3 +37,10 @@ jobs:
         with:
           name: Documentation
           path: doxygen
+          
+      - name: Deploy to GitHub Pages
+        uses: JamesIves/github-pages-deploy-action@3.7.1
+        with:
+          github_token: ${{ github.token }}
+          branch: gh-pages
+          folder: doxygen/html
```
JanosGit, line 47: Since this workflow is triggered on every pull request to develop and push to master and develop, this action will be run on each event. This means that depending on if the last trigger was to the master or the develop branch the GitHub pages will show the master or develop documentation? And if the last trigger was a pull request it will temporarily view the state of the pull request until it has been merged?

I actually had a solution in mind like e.g. JUCE offers where there are always two versions of the documentation available, the one of the latest develop tip and the one of the latest master release.

If we want to keep it simple, I think having a solution where we only deploy on pushes to develop would be fine for now, given that we haven't released anything to the master so far anyway.

JanosGit (changes requested):

.github/workflows/Documentation.yml
psciri: so we should rather extract the entire doc step to a separate action?

psciri: or restrict _this_ action?

JanosGit: Uhm the documentation step is a separate action workflow decoupled from the build & unit test currently

.github/workflows/Documentation.yml
```diff
@@ -37,3 +37,11 @@ jobs:
         with:
           name: Documentation
           path: doxygen
+          
+      - name: Deploy to GitHub Pages (if on develop or main branch)
+        if: ${{ github.ref_name == 'main' || github.ref_name == 'develop' }}
```
psciri: @JanosGit this should omit deployments of PRs

JanosGit: I'd actually only do this for the `develop` branch. Otherwise the same situation occurs, the documentation either reflects the `main` or `develop` depending on what has been merged last

.github/workflows/Documentation.yml
```diff
@@ -37,3 +37,11 @@ jobs:
         with:
           name: Documentation
           path: doxygen
+          
+      - name: Deploy to GitHub Pages (if on develop or main branch)
+        if: ${{ github.ref_name == 'develop' }}
```
psciri, line 42: @JanosGit like so?

JanosGit: I like!

JanosGit (approved):

PR #123, opened 2023-10-23
Expressions: Add target attributes to SIMD evaluation preparation calls

### LINKED ISSUE #121: Update to recent version breaks division operator (Ubuntu, gcc)
Updating to develop tip seems to make division operator not play nicely anymore. Consider the following code:
```
vctr::Array<float, 2> foo{1.0f, 2.0f};
foo /= 2.0f;

for (auto& val : foo)
    DBG (val);
```
The output is
```
inf
inf
```

Ubuntu 20.04
gcc 11.4.0

Happy to provide more details if needed :slightly_smiling_face:

JanosGit:
This fixes issues when building with GCC 12 where broadcast instructions were not executed as expected

### DIFF
--- a/include/vctr/Expressions/BasicMath/Add.h
+++ b/include/vctr/Expressions/BasicMath/Add.h
@@ -108,8 +108,15 @@ public:
 
     //==============================================================================
     // AVX Implementation
-    void prepareAVXEvaluation() const
-    requires has::prepareAVXEvaluation<SrcType>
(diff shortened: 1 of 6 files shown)


JanosGit:
Seems there is now some completely conan related problem with the Windows builds...

TheWaveWarden (approved):
Code looks good, although I won't be able to verify the fix on my system before friday!

PR #124, opened 2023-11-02
Array: Fixed error for deleted assignment operator element types. Make sure that Array ctor is only exectued for copy constructible element types.

JanosGit:
…e sure that Array ctor is only exectued for copy constructible element types.


p4th4x0r (approved):
Fine,

PR #125, opened 2023-12-20
Vctr 87 enhance documentation and examples to give the user a good overview of what is possible with vctr

JanosGit:


demo/demo.cpp
```diff
@@ -77,7 +77,38 @@ int main()
     // Looks good!
     std::cout << "x = \t[ " << x[0] << ", " << x[1] << ", " << x[2] << ", " << x[3] << " ]" << std::endl;
 
+    // We can transform the content of container by applying a function to each element. In this case,
```
iromur: of **a** container?

include/vctr/Miscellaneous/StdOstreamOperator.h
```diff
@@ -26,7 +26,7 @@ namespace vctr
 template <is::anyVctrOrExpression V>
 std::ostream& operator<< (std::ostream& s, const V& vec)
 {
-    if (vec.empty())
+    if (vec.size() == 0)
```
iromur, line 29: Just out if interest: Why?

iromur (approved):

demo/demo.cpp
JanosGit: Sure

include/vctr/Miscellaneous/StdOstreamOperator.h
JanosGit: since expressions are only required to implement a size member function, so this didn't work when attempting to print an expression result

PR #127, opened 2024-03-28
Unit Tests: Fixed compile error in Divide test with newer Apple Clang

JanosGit:
I ran into a compile error here when attempting to build the unit tests with a recent Xcode version. Seems like the compiler attempted to instantiate the function template with some invalid types. The lambdas with explicit argument types fixed it.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -33,7 +33,7 @@ jobs:
         run: doxygen-1.9.6/bin/doxygen
 
       - name: Save Artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: Documentation
           path: doxygen
--- a/.github/workflows/UnitTests.yml
+++ b/.github/workflows/UnitTests.yml
@@ -17,9 +17,9 @@ jobs:
           #  cxx: clang++-15
           #  ipp: no_ipp
 
-          - runner: ubuntu-22.04
-            cc: gcc-12
-            cxx: g++-12
+          - runner: ubuntu-24.04
+            cc: gcc-13
+            cxx: g++-13
             ipp: no_ipp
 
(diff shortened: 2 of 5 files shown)


iromur (approved):
That's really strange. Aren't other arithmetic tests done in the same way?

Another question: Are the unit test failures on Linux related or for another time?

JanosGit:
From what I get this is caused by some explicitly overloaded `operator/` definitions from the Xcode 15 standard library. Unfortunately, the change which fixes the macOS build now explicitly breaks the Linux build.

JanosGit:
I resurrected this PR again after realizing that the GitHub hosted macOS 12 runner that we used is no longer available and thus the problem solved in the original PR was also relevant for Actions builds now.

So I first updated the UnitTest workflow to use the new runners, which also finally added an ARM based mac runner, and while being there, adding a compile definition to silence compile warnings regarding a new LAPACK version in macOS builds.

As the changes to the unit test code broke the linux build, I then tried the path of least resistance and used a newer GCC version which fixed that problem. However, updating that toolchain resulted in a build error in catch2 which could be resolved by updating that dependency to a newer version.

I could however not update catch2 to the latest version because that package is Conan 2 only. So I think the next step should be a migration to Conan 2 here to be able to get the latest packages in future.

iromur:
Thanks for the detailed explanation. All very reasonable, code looks fine!
A pity for the gcc update but I guess we are not much into keeping compatibility with different gcc versions anyway, do we?
Might become relevant when this is going to be used in less bleeding-edge systems (e.g., some embedded platforms use older compilers that are not easily updated etc.), but in the end C++-20 is not for historic platforms anyway.

iromur (approved):

JanosGit:
Yeah, regarding gcc compatibility note that the issue was the unit test code, not the project code itself. Chances are high that the actual vctr implementation will still compile fine with older GCC versions. I just took the path of least resistance making the unit tests compile again, given that I had no Linux machine at hand for easy local reproduction of the issue. We can still add an issue to try to get the unit tests building with older GCC versions again.

PR #130, opened 2024-05-29
Expressions: Make deduction of value_type for binary expressions more robust

JanosGit:
… robust

### DIFF
--- a/include/vctr/Expressions/BasicMath/Add.h
+++ b/include/vctr/Expressions/BasicMath/Add.h
@@ -28,12 +28,15 @@ namespace vctr::expressions
 template <size_t extent, class SrcAType, class SrcBType>
 class AddVectors : ExpressionTemplateBase
 {
+private:
+    static auto op (const ValueType<SrcAType>& a, const ValueType<SrcBType>& b) { return a + b; }
+
 public:
-    using value_type = std::common_type_t<ValueType<SrcAType>, ValueType<SrcBType>>;
+    using value_type = CommonVecExpressionType<ValueType<SrcAType>, ValueType<SrcBType>, op>;
 
     VCTR_COMMON_BINARY_VEC_VEC_EXPRESSION_MEMBERS (AddVectors, srcA, srcB)
 
-    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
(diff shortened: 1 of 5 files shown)


include/vctr/Expressions/BasicMath/Add.h
```diff
@@ -29,11 +29,11 @@ template <size_t extent, class SrcAType, class SrcBType>
 class AddVectors : ExpressionTemplateBase
 {
 public:
-    using value_type = std::common_type_t<ValueType<SrcAType>, ValueType<SrcBType>>;
+    using value_type = CommonVecExpressionType<ValueType<SrcAType>, ValueType<SrcBType>, [] (auto&& a, auto&& b) { return a + b; }>;
```
TheWaveWarden: I can reproduce the compile errors locally:
```
C:\Users\frederik.siepe\cleanvoice\ext\pago-modules\ext\VCTR\include\vctr\Expressions\BasicMath\Add.h(32,130): error C2888:
 'auto <lambda_3_>::operator ()(_T1 &&,_T2 &&) const': symbol cannot be defined within namespace 'expressions' [C:\Users\fr
ederik.siepe\cleanvoice\build\cleanvoice.vcxproj]
  (compiling source file '../src/plugin/processor/offline_analysis/deesser/realtime_inference/HybridModel.cpp')
  C:\Users\frederik.siepe\cleanvoice\ext\pago-modules\ext\VCTR\include\vctr\Expressions\BasicMath\Add.h(32,130):
  the template instantiation context (the oldest one first) is
        C:\Users\frederik.siepe\cleanvoice\ext\pago-modules\ext\VCTR\include\vctr\Expressions\BasicMath\Add.h(29,7):
        while compiling class template 'vctr::expressions::AddVectors'
```
The compiler seems to get hung up on the fact that we're in namespace `expressions` and it wants the lambda (?) to be defined in namespace `detail`. I played around a bit, but couldn't make it work so far...

https://learn.microsoft.com/de-de/cpp/error-messages/compiler-errors-2/compiler-error-c2888?view=msvc-170

JanosGit: I reworked it by using a static member function instead of a lambda now. Let's see if that works...

TheWaveWarden (approved):

PR #131, opened 2024-07-10
Vctr 125 add unit test for span cast to

JanosGit:


include/vctr/Containers/Span.h
```diff
@@ -189,7 +189,7 @@ class Span : public VctrBase<ElementType, std::span<ElementType, extent>, extent
     template <class DstType>
     constexpr auto castTo()
     {
-        using SizeRatio = std::ratio<sizeof (DstType) / sizeof (ElementType)>;
+        using SizeRatio = std::ratio<sizeof (ElementType), sizeof (DstType)>;
```
JanosGit, line 205: Yep, this was double wrong. We did not only switch numerator and denominator but also didn't pass them correctly to `std::ratio` as comma seperated arguments but as a fraction. Since `std::ratio` has a default value of 1 defined for the denominator this lead to an integer division being performed and the result being applied as numerator. Took me quite some time to spot this...

iromur: 🙈

test/TestCases/SpanMemberFunctions.cpp
```diff
@@ -44,3 +44,38 @@ TEST_CASE ("Assignment operator", "[SpanMemberFunctions]")
...
+    {
+        constexpr vctr::Array<int32_t, 32> a { 0 };
+
+        auto asInt16Fixed = vctr::Span (a).castTo<const int16_t>();
+        auto asInt64Fixed = vctr::Span (a).castTo<const int64_t>();
+
+        STATIC_REQUIRE (asInt16Fixed.size() == 64);
+        STATIC_REQUIRE (asInt64Fixed.size() == 16);
+
+        REQUIRE (asInt16Fixed.allElementsEqual (0));
```
iromur: Given that the original array has only 0 as members, I am not sure whether this actually tests anything substantial, except of no flipped bits are getting introduced.
Wouldn't it make more sense to initialise `a` to `1 << 16 | 1` and compare against `1` (int16 case) or `1 << 48  |  1 << 32 | 1 << 16 | 1` (64bit case)?

iromur (approved):
Looks good besides minor comment!

test/TestCases/SpanMemberFunctions.cpp
JanosGit: Yes, that's a clever idea, will do that

PR #133, opened 2024-09-03
Span: Added implicit conversion operator to std::span

JanosGit:
This was already missing for far too long ;)


p4th4x0r (approved):
Approved

PR #136, opened 2025-02-18
Expressions: Added RealToComplex

JanosGit:


include/vctr_utils/vctr_juce_helpers.h
```diff
@@ -56,26 +56,26 @@ namespace vctr::juce_helpers
     @see VctrBase::getAVX
  */
 template <is::anyVctr Src>
-auto toSIMDRegister (const Src& src)
+auto toSIMDRegister (const Src& src, size_t i = 0)
```
pmglr, line 63: Could you please document the second parameter?

Also, although kind of obvious from the method name, I'd like formal documentation of the return value since we are using `auto` here.

pmglr: > Loads the first elements of src into a juce::SIMDRegister.

This needs to be reformulated

test/TestCases/Expressions/RealImag.cpp
```diff
@@ -51,11 +51,28 @@ TEMPLATE_PRODUCT_TEST_CASE ("Imag", "[complex]", (PlatformVectorOps, VCTR_NATIVE
...
+{
+    VCTR_TEST_DEFINES (10)
+
+    const vctr::Vector cplx = filter << vctr::realToComplex << srcA;
+    const vctr::Vector cplxU = filter << vctr::realToComplex << srcUnaligned;
+    
+    static_assert (vctr::is::complexFloatNumber<typename decltype (cplx)::value_type>);
+    static_assert (vctr::is::complexFloatNumber<typename decltype (cplxU)::value_type>);
+
+    REQUIRE_THAT (srcA, vctr::EqualsTransformedBy<getRealPart> (cplx));
```
pmglr, line 73: Those vctr matchers are more powerful than I thought!

test/TestCases/Expressions/RealImag.cpp
```diff
@@ -51,11 +51,28 @@ TEMPLATE_PRODUCT_TEST_CASE ("Imag", "[complex]", (PlatformVectorOps, VCTR_NATIVE
...
     REQUIRE_THAT (imag, vctr::EqualsTransformedBy<getImagPart> (srcA));
     REQUIRE_THAT (imagU, vctr::EqualsTransformedBy<getImagPart> (srcUnaligned));
+}
+
+TEMPLATE_PRODUCT_TEST_CASE ("RealToComplex", "[complex]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double) )
+{
+    VCTR_TEST_DEFINES (10)
+
+    const vctr::Vector cplx = filter << vctr::realToComplex << srcA;
+    const vctr::Vector cplxU = filter << vctr::realToComplex << srcUnaligned;
```
pmglr, line 68: I assume this creates real complex vectors? If that is so, maybe let's also test with data with imag != 0?

JanosGit: The new expression converts a source `Vector<float>` into a destination `Vector<std::complex<float>>` with all imaginary parts in the destination vector being zero. I'm not exactly sure what you mean by "test with data with imag != 0?" when the source is a real valued float vector?

pmglr: Sorry, I was momentarily confused and thought we are generating test data for a complex-to-real expression further below while this is already the expression under test. You can ignore my comment.

include/vctr_utils/vctr_juce_helpers.h
JanosGit: Thx, I updated the comment

pmglr (approved):

PR #138, opened 2025-03-12
Updated project to be compatible with conan2

TheWaveWarden:


cmake/FindIPP.cmake
JanosGit, line 1: What's the background of these changes? Seems pretty unrelated to conan changes on the first glance?

conanfile.txt
```diff
@@ -0,0 +1,3 @@
+[requires]
+catch2/3.7.0
```
JanosGit: Is this the latest catch release on conan center?

TheWaveWarden: Nope, this is the version we used beforehand. I'll update it

cmake/FindIPP.cmake
TheWaveWarden: I'm not completely sure TBH. See the builds for the second commit, where builds would fail claiming that the target `ipp` already exists. I tried this to circumnavigate it and the runners seemed to be happy..

conanfile.txt
TheWaveWarden: Ok it's at the latest 3.8.0 now

cmake/FindIPP.cmake
TheWaveWarden: I've tested around a bit and in a trace I could see that these multiple calls indeed are coming from within the cmake-conan package. My best guess is that it's supposed to be this way and perhaps adding the target inside the `findX.cmake` script is non-standard or we have to make sure we don't add twice (like we now do)? Sorry I don't have a complete/satisfying answer here..

JanosGit: I see. I'd propose to split this out into a separate issue and get it merged in this state for now.

JanosGit (approved):

cmake/FindIPP.cmake
TheWaveWarden: Ok, I added an issue in this Repo!

PR #140, opened 2025-06-03
Expressions: Added Apple Accelerate overload for complex multiply-add evaluation

JanosGit:


p4th4x0r (approved):
This was what i was thinking about after the stand-up :)

PR #143, opened 2025-09-17
Vctr 149 add SIMD based fast math approximation functions

TheWaveWarden:


include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -0,0 +1,175 @@
...
+
+        auto denominator = Expression::Neon::add (src.getNeon (i), SIMDConstMinus20.neon);
+        denominator = Expression::Neon::mul (denominator, src.getNeon (i));
+        denominator = Expression::Neon::add (denominator, SIMDConst180.neon);
+        denominator = Expression::Neon::mul (denominator, src.getNeon (i));
+        denominator = Expression::Neon::add (denominator, SIMDConstMinus840.neon);
+        denominator = Expression::Neon::mul (denominator, src.getNeon (i));
+        denominator = Expression::Neon::add (denominator, SIMDConst1680.neon);
+
+        // Neon doesn't implement division, so we need to do it manually
```
TheWaveWarden: outdated comment

include/vctr/Expressions/BasicMath/Add.h
```diff
@@ -163,6 +163,70 @@ class AddSingleToVec : ExpressionTemplateBase
...
+        src.prepareSSEEvaluation();
+        constantSIMD.sse = Expression::SSE::broadcast (constant);
+    }
+
+    VCTR_FORCEDINLINE VCTR_TARGET ("sse4.1") SSERegister<value_type> getSSE (size_t i) const
+    requires (archX64 && has::getSSE<SrcType> && Expression::allElementTypesSame && Expression::CommonElement::isRealFloat)
+    {
+        return Expression::SSE::add (constantSIMD.sse, src.getSSE (i));
+    }
+
```
JanosGit, line 225: I guess you didn't add full blown X64 integer and ARM Neon implementations as you didn't actually need this in the end? I'm fine with that, just wonder if we should start logging these gaps and create an overview of which implementation is supplied for which expression

include/vctr/Expressions/BasicMath/Add.h
```diff
@@ -201,4 +265,11 @@ constexpr auto operator+ (Src&& vec, typename std::remove_cvref_t<Src>::value_ty
     return expressions::AddSingleToVec<extentOf<Src>, Src> (single, std::forward<Src> (vec));
 }
 
+/** Returns an expression that adds a compile time constant to a vector or expression source.
+
+    @ingroup Expressions
+ */
+template <auto constantValue>
+constexpr inline ExpressionChainBuilder<expressions::AddConstantToVec, Constant<constantValue>> addConstant;
```
JanosGit, line 273: As we have tests for all expressions in place, please add a small test case for this expression to the Add test

include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -0,0 +1,175 @@
...
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+namespace vctr::expressions
+{
+
+//==============================================================================
+/** Calculates a fast approximation for the exp function. */
+template <size_t extent, class SrcType>
```
JanosGit, line 28: This should probably be constrained to std::floating_point value types in the source like

```
requires std::floating_point<ValueType<SrcType>>
```

include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -0,0 +1,175 @@
...
+class FastExp : ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (FastExp, src)
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        return (value_type (1680) + src[i] * (value_type (840) + src[i] * (value_type (180) + src[i] * (value_type (20) + src[i])))) / (value_type (1680) + src[i] * (value_type (-840) + src[i] * (value_type (180) + src[i] * (value_type (-20) + src[i]))));
```
JanosGit: Should we maybe make these constants `static constexpr value_type` members?

include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -0,0 +1,175 @@
...
+        auto denominator = Expression::SSE::add (src.getSSE (i), SIMDConstMinus20.sse);
+        denominator = Expression::SSE::mul (denominator, src.getSSE (i));
+        denominator = Expression::SSE::add (denominator, SIMDConst180.sse);
+        denominator = Expression::SSE::mul (denominator, src.getSSE (i));
+        denominator = Expression::SSE::add (denominator, SIMDConstMinus840.sse);
+        denominator = Expression::SSE::mul (denominator, src.getSSE (i));
+        denominator = Expression::SSE::add (denominator, SIMDConst1680.sse);
+
+        return Expression::SSE::div (numerator, denominator);
+    }
```
JanosGit, line 114: Just for the sake of consistency, I would place the SSE implementation before the Neon implementation

test/TestCases/Expressions/FastExp.cpp
```diff
@@ -0,0 +1,52 @@
...
+    version 3 along with VCTR.  If not, see <https://www.gnu.org/licenses/>.
+  ==============================================================================
+*/
+
+#include <vctr_test_utils/vctr_test_common.h>
+
+template <vctr::is::realOrComplexFloatNumber T>
+T stdExp (T v) { return std::exp (v); }
+
+template <vctr::is::realOrComplexFloatNumber T>
```
JanosGit, line 28: oh so this works for complex numbers as well?

JanosGit (commented):
Nice Work! Would second the AI regarding the commented out test cases, probably a leftover to make it build with msvc locally?

include/vctr/Expressions/BasicMath/Add.h
TheWaveWarden: Yeah, thought the same, it would be nice to have an overview and then maybe close them bit by bit.
This implementation was taken / adjusted from another one, I think Multiply?

test/TestCases/Expressions/FastExp.cpp
TheWaveWarden: The SIMD implementations of FastExp limit it to floating types. However, when I change it to float here, the term `EqualsTransformedBy` fails to compile. I think since the simd implementation is for floating points only, so the complex test fall back to the sample wise calculation, which should work regardless.

include/vctr/Expressions/BasicMath/Add.h
JanosGit: Yeah, let's leave it like that for now. I'll add a task to the backlog to create a full overview of the missing pieces as a next step

test/TestCases/Expressions/FastExp.cpp
JanosGit: Yeah that's to be expected, I just was surprised that the appxomiation calculation itself also works for complex numbers - which is actually probably not even that surprising thinking about it twice ;)

In that case I'd change my comment on constraining the actual expression to floating point and would now suggest to constrain it using is::realOrComplexFloatNumber

include/vctr/Expressions/BasicMath/Add.h
TheWaveWarden: done! (is there a way to open tasks here?)

include/vctr/Expressions/DSP/FastExp.h
TheWaveWarden: done

include/vctr/Expressions/DSP/FastExp.h
TheWaveWarden: done

test/TestCases/Expressions/FastExp.cpp
TheWaveWarden: I think the approximation should even be reasonable in the complex plane around 0.

test/TestCases/Expressions/Add.cpp
```diff
@@ -48,12 +48,15 @@ TEMPLATE_PRODUCT_TEST_CASE ("Addition in place", "[VCTR][add]", (PlatformVectorO
 
     auto sum1 = srcA;
     auto sum2 = srcA;
+    auto sum3 = srcA;
 
     sum1 += srcB;
     sum2 += c;
+    sum3 = vctr::addConstant<ElementType (42)> << srcA;
```
JanosGit: You put this in the "Addition in place" section which seems to not really fit to me ;)

test/TestCases/Expressions/Add.cpp
```diff
@@ -48,6 +50,7 @@ TEMPLATE_PRODUCT_TEST_CASE ("Addition in place", "[VCTR][add]", (PlatformVectorO
 
     auto sum1 = srcA;
     auto sum2 = srcA;
+    auto sum3 = srcA;
```
JanosGit: this one is probably no longer needed now ;)

JanosGit (approved):

PR #146, opened 2025-10-08
Implement SIMD Bitmasks

### LINKED ISSUE #86: Installing llvm 15 fails on Linux runner
clang 15 based Linux runners have temporarily been disabled

TheWaveWarden:


include/vctr/Expressions/BasicMath/CompareMask.h
```diff
@@ -0,0 +1,159 @@
...
+
+    //VCTR_FORCEDINLINE VCTR_TARGET ("avx") AVXRegister<value_type> getAVX (size_t i) const
+    //{
+    //    const __m256 compareResult = _mm256_cmp_ps (a.getAVX (i).value, b.getAVX (i).value, int (Op));
+    //    const int mask = _mm256_movemask_ps (compareResult);
+    //
+    //    // write bool output vector
+    //    for (unsigned i = 0; i < lanes; ++i)
+    //        dst[registerIndex + i] = ((mask >> i) & 1) != 0;
+    //}
```
TheWaveWarden: Ok, here the pattern I used before with `getAVX()` didn't work out, because the return type is not the same as the inputs? I used the `evalNextVectorOpInExpressionChain()` below instead to run over the entire buffer

include/vctr/Expressions/BasicMath/CompareMask.h
```diff
@@ -0,0 +1,159 @@
...
+        for (size_t r = 0; r < regCount; ++r)
+        {
+            const auto registerIndex = r * lanes;
+
+            const auto compareResult = _mm256_cmp_ps (a.getAVX (registerIndex).value, b.getAVX (registerIndex).value, int (Op));
+            const auto mask = _mm256_movemask_ps (compareResult);
+
+            // write bool output vector from bitmask
+            for (size_t i = 0; i < lanes; ++i)
+                dst[registerIndex + i] = ((mask >> i) & 1) != 0;
```
TheWaveWarden: Here' we're doing an extra loop to pack the bitmask back into the bool vector. This seems like it breaks the entire SIMD idea, because then we could also just write a loop which compares the inputs to begin with...

JanosGit:
I'm not really sure why the linux build segfaults. Don't have a linux machine at hand to test it locally right now. Do you see anything obvious that could be wrong @pmglr?

pmglr:
> I'm not really sure why the linux build segfaults. Don't have a linux machine at hand to test it locally right now. Do you see anything obvious that could be wrong @pmglr?

Did it already fail before your changes or only after?

include/vctr/SIMD/AVX/AVXRegister.h
```diff
@@ -90,9 +97,16 @@ struct AVXRegister<double>
...
 
+    //==============================================================================
+    // Generate Compare Masks
+    template <CompareOp Op>
+    VCTR_TARGET ("avx") static AVXRegister compareMask (AVXRegister a, AVXRegister b) { return { _mm256_cmp_pd (a.value, b.value, int (Op)) }; }
+
     //==============================================================================
     // Bit Operations
-    VCTR_TARGET ("avx") static AVXRegister andNot (AVXRegister a, AVXRegister b) { return { _mm256_andnot_pd (a.value, b.value) }; }
+    VCTR_TARGET ("avx") static AVXRegister bitwiseAndNot (AVXRegister a, AVXRegister b) { return { _mm256_andnot_pd (a.value, b.value) }; }
```
pmglr: How about `bitwiseNand`?

pmglr: ok, doesn't seem to be a common way to phrase it if I look at the different SIMD instructions

JanosGit:
> > I'm not really sure why the linux build segfaults. Don't have a linux machine at hand to test it locally right now. Do you see anything obvious that could be wrong @pmglr?
> 
> Did it already fail before your changes or only after?

Only after my changes. Let's see if @iromur finds out something when debugging it locally

test/TestCases/BitwiseMaskOperations.cpp
```diff
@@ -0,0 +1,157 @@
...
+using RegisterType = vctr::NeonRegister<T>;
+#else
+template <class T>
+using RegisterType = vctr::AVXRegister<T>;
+#if VCTR_GCC
+// In contrast to other test cases where we test higher level VCTR functions, we call register member functions directly here.
+// All the AVX functions are attributed with __attribute__ ((target ("avx"))). Directly taking the return value of those functions
+// here leads to issues with GCC obviously not using the correct calling conventions when compiling the test case without explicitly
+// setting the AVX target. This pragma enables it for the entire translation unit
+#pragma GCC target ("avx")
```
JanosGit, line 84: This finally helped fixing the GCC based linux crashes. There was a build warning "The ABI for passing parameters with 32-byte alignment has changed in GCC 4.6" that helped me understanding the issue as it lead me to this SO post which has all the details https://stackoverflow.com/questions/39383193/compiling-legacy-gcc-code-with-avx-vector-warnings

.github/workflows/UnitTests.yml
```diff
@@ -11,22 +11,20 @@ jobs:
...
           - runner: ubuntu-24.04
             cc: gcc-13
             cxx: g++-13
             ipp: no_ipp
 
-          # Linux clang 15 runners are temporarily disabled due to installation issues. See https://github.com/sonible/VCTR/issues/86
-          # - runner: ubuntu-22.04
-          #  cc: clang-15
-          #  cxx: clang++-15
-          #  ipp: ipp
```
JanosGit, line 29: Found out that the clang installation issues have been fixed in the meantime as clang-15 is now pre-installed on the ubuntu runners, so finally solved that old issue along the way

test/TestCases/BitwiseMaskOperations.cpp
iromur: Glad you found it, congrats! 🥇 Can confirm that the crash at the place we investigated is gone also locally.

Locally, I now have this test failure:
```
SIMDFilter - AnySIMD<float>
-------------------------------------------------------------------------------
/home/rumori/src/VCTR/test/TestCases/Expressions/SIMDFilter.cpp:28
...............................................................................

/home/rumori/src/VCTR/test/TestCases/Expressions/SIMDFilter.cpp:28: FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
```

Turns out that this is caused by:
```
frame #4: 0x000055555596da01 vctr_test`vctr::expressions::AssertSIMDFilter<10ul, vctr::expressions::AnySIMDFilter<10ul, vctr::expressions::AddVectors<10ul, vctr::Array<float, 10ul, 16ul> const&, vctr::Array<float, 10ul, 16ul> const&>>>::operator[](this=0x00007fffffffd690, i=0) const at SIMDFilter.h:152:9
   149 	
   150 	   VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
   151 	   {
-> 152 	       VCTR_ASSERT (false);
   153 	       return src[i];
   154 	   }
   155 	
```
Is `VCTR_ASSERT (false)` intended here? (Also a bit below in line 166 in the same header?)

Locally, I have gcc 15.2.1.
Also still have the narrowing warnings but probably a different issue.

JanosGit: That's interesting. The assertion that you hitting is actually intended in that filter expression but that it's hit is unintended and that it aborts the application is also unintended.

The docs of `vctr::assertSIMD` explain the use-case of the expression:
```
/** This filter inserts an assertion that will be hit in case that an expression is evaluated with anything but SIMD operations.

    This is mainly helpful to ensure that your explicit SIMD based expressions do not accidentally run differently.

    @ingroup Expressions
 */
```

So here it should ensure that we are actually testing a SIMD evaluation and it turns out, that on that platform the implementation falls back to for loop evaluation which is not expected in that case. This probably didn't come up until now as the pipeline only tests release builds where this assertion is not active. So there is an issue here on Linux/gcc that we should look at. However, that seems to have been around for a while, we just didn't run any debug build. I think that this could be improved by reworking `vctr::assertSIMD` in a way that we can e.g. pass a lambda being executed as argument to the expression and pass a catch `FAIL` expression in there so that this condition can also be discovered in regular test runs.

Another thing that is surprising to me is, that the assertion aborts the application. The `VCTR_ASSERT` macro is intended to stop under the debugger but allow the user to continue execution. Seems that at least for Linux/gcc this is not the case.

And regarding the narrowing warnings, these should be fixed.

I added issues for all three:
https://github.com/sonible/VCTR/issues/148
https://github.com/sonible/VCTR/issues/149
https://github.com/sonible/VCTR/issues/150

So to conclude, we have ssues here that we should look at, but none of these are related to the changes from this Pull Request, so I'd propose merging these changes even with those issues being present.

iromur (approved):
Approved that crashes on Linux are solved and issues stumbled across locally went into follow-up tasks.

include/vctr/SIMD/AVX/AVXRegister.h
```diff
@@ -90,9 +97,16 @@ struct AVXRegister<double>
     VCTR_TARGET ("avx") void storeUnaligned (double* d) const { _mm256_storeu_pd (d, value); }
     VCTR_TARGET ("avx") void storeAligned   (double* d) const { _mm256_store_pd (d, value); }
 
+    //==============================================================================
+    // Generate Compare Masks
+    template <CompareOp Op>
+    VCTR_TARGET ("avx") static AVXRegister compareMask (AVXRegister a, AVXRegister b) { return { _mm256_cmp_pd (a.value, b.value, int (Op)) }; }
```
pmglr: I'm not sure if maybe the name can be misleading (as in comparing a mask). `getCompareMask`?

include/vctr/SIMD/Neon/NeonRegister.h
```diff
@@ -31,6 +31,28 @@ struct NeonRegister
...
+template <> struct NeonCompare<CompareOp::greater, double>        { static auto cmp (float64x2_t a, float64x2_t b) { return vcgtq_f64 (a, b); } };
+template <> struct NeonCompare<CompareOp::greaterOrEqual, float>  { static auto cmp (float32x4_t a, float32x4_t b) { return vcgeq_f32 (a, b); } };
+template <> struct NeonCompare<CompareOp::greaterOrEqual, double> { static auto cmp (float64x2_t a, float64x2_t b) { return vcgeq_f64 (a, b); } };
+template <> struct NeonCompare<CompareOp::equal, float>           { static auto cmp (float32x4_t a, float32x4_t b) { return vceqq_f32 (a, b); } };
+template <> struct NeonCompare<CompareOp::equal, double>          { static auto cmp (float64x2_t a, float64x2_t b) { return vceqq_f64 (a, b); } };
+template <> struct NeonCompare<CompareOp::notEqual, float>        { static auto cmp (float32x4_t a, float32x4_t b) { return vmvnq_u32 (vceqq_f32 (a, b)); } };
+template <> struct NeonCompare<CompareOp::notEqual, double>       { static auto cmp (float64x2_t a, float64x2_t b) { return vreinterpretq_u64_u32 (vmvnq_u32 (vreinterpretq_u32_u64 (vceqq_f64 (a, b)))); } };
```
iromur, line 52: Pff. So there is no `vmvnq_u64` apparently?

include/vctr/SIMD/SIMDHelpers.h
```diff
@@ -45,6 +45,30 @@ constexpr size_t nextMultipleOf (size_t numElements)
 namespace vctr
 {
 
+#if (! VCTR_X64)
+// These are defined in immintrin.h – defining them here for non x86 builds
+// to be able to assign the CompareOp enum cross-platform
+// NOLINTBEGIN (*-reserved-identifier)
+constexpr auto _CMP_LT_OQ = 0;
+constexpr auto _CMP_LE_OQ = 1;
+constexpr auto  _CMP_GT_OQ = 2;
```
iromur: Looks like an additional blank before _CMP?

test/TestCases/BitwiseMaskOperations.cpp
```diff
@@ -0,0 +1,157 @@
...
+        const auto b = bSrc.getNeon (0);
+#else
+        const auto a = aSrc.getAVX (0);
+        const auto b = bSrc.getAVX (0);
+#endif
+        const auto mask = RegisterType<T>::template compareMask<registerCompareOp> (a, b);
+
+        // perform other bitwise operations
+        const auto bitAnd = RegisterType<T>::bitwiseAnd (a, mask);
+        const auto bitAndNot = RegisterType<T>::bitwiseAndNot (mask, a); // careful: First argument is negated, not the second one!
```
iromur: Should that convention be documented somewhere in the API rather than the test code?
As it looks like, this convention follows that of the AVX/SSE instructions?

iromur (approved):
Looks good to me!

include/vctr/SIMD/AVX/AVXRegister.h
JanosGit: I get your point, not super happy with that too. I try to keep the names in this class reasonably short here as this domain is full of abbreviations anyway and in case you write SIMD code you are usually a bit fluent with those terms and what to expect behind them and otherwise code that invokes those function becomes quite lengthy. So what about a plain `compare`? I think it's obvious that this results in a bit mask in the context of SIMD code

include/vctr/SIMD/Neon/NeonRegister.h
JanosGit: For whatever reason, no... https://developer.arm.com/architectures/instruction-sets/intrinsics/#f:@navigationhierarchiessimdisa=[Neon]&f:@navigationhierarchiesarchitectures=[A64]&q=vmvnq_u64

test/TestCases/BitwiseMaskOperations.cpp
JanosGit: This was the result of refactoring an already existing implementation. But I also wondered if we should just swap the arguments because that seems more intuitive to me. Now that you are commenting on this too, I think I'll do that. But it's probably still worth documenting this even better.

pmglr (approved):

PR #151, opened 2025-11-04
Vctr 152 implement fast db calculations

JanosGit:
This ended up to be quite a lot of work, but it seems worth it. When compiling with clang, speedups are something between factor 10 to factor 5, depending on the platform. With MSVC I experienced at least a speedup of factor 2. This takes me to the conclusion that
a) a general compiler switch to clang on Windows might really be a good idea in general
b) we should add an extensive benchmark suite to this project finally
c) we should investigate why code generated with MSVC performs that worse compared with code generated with clang

### DIFF
--- a/benchmark/benchmark.cpp
+++ b/benchmark/benchmark.cpp
@@ -30,6 +30,7 @@ int main (int argc, char** argv) { return Catch::Session().run (argc, argv); }
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/generators/catch_generators.hpp>
 #include <chrono>
+#include <vctr_test_utils/vctr_test_common.h>
 
 // ===============================================================
 TEST_CASE ("Decibels")
@@ -187,4 +188,30 @@ TEST_CASE ("Multiplication")
             return notAccelerated();
         };
     }
+}
+
+TEST_CASE ("Precise vs fast decibel benchmarks",  "[VCTR][Decibels]")
+{
+    auto srcMag16 = UnitTestValues<float>::array<16, 0, 0, 2> (false, true);
+    auto srcDb16 = UnitTestValues<float>::array<16, 0, -100, 10> (false, true);
+
+    BENCHMARK ("Precise magToDb")
+    {
+        return vctr::Array (vctr::magToDb<vctr::dBFS> << srcMag16);
+    };
+
+    BENCHMARK ("Fast magToDb")
+    {
+        return vctr::Array (vctr::fastMagToDb<vctr::dBFS> << srcMag16);
+    };
+
+    BENCHMARK ("Precise dbToMag")
+    {
+        return vctr::Array (vctr::dbToMag<vctr::dBFS> << srcDb16);
+    };
+
+    BENCHMARK ("Fast dbToMag")
+    {
+        return vctr::Array (vctr::fastDbToMag<vctr::dBFS> << srcDb16);
+    };
 }
--- a/include/vctr/Expressions/BasicMath/Clamp.h
+++ b/include/vctr/Expressions/BasicMath/Clamp.h
@@ -147,6 +147,90 @@ public:
     }
 
     //==============================================================================
+    // AVX Implementation
+    VCTR_FORCEDINLINE VCTR_TARGET ("avx") void prepareAVXEvaluation() const
+    requires (has::prepareAVXEvaluation<SrcType> && Expression::CommonElement::isRealFloat)
+    {
+        src.prepareAVXEvaluation();
+
+        if constexpr (clampLow)
+            lowerBoundSIMD.avx = Expression::AVX::broadcast (lowerBound);
+
+        if constexpr (clampHigh)
+            upperBoundSIMD.avx = Expression::AVX::broadcast (upperBound);
+    }
+
+    VCTR_FORCEDINLINE VCTR_TARGET ("fma") AVXRegister<value_type> getAVX (size_t i) const
+    requires (archX64 && has::getAVX<SrcType> && Expression::allElementTypesSame && Expression::CommonElement::isRealFloat)
+    {
+        auto x = src.getAVX (i);
+
+        if constexpr (clampLow)
+            x = Expression::AVX::max (lowerBoundSIMD.avx, x);
+
+        if constexpr (clampHigh)
+            x = Expression::AVX::min (x, upperBoundSIMD.avx);
+
+        return x;
+    }
+
+    //==============================================================================
+    // SSE Implementation
+    VCTR_FORCEDINLINE VCTR_TARGET ("sse4.1") void prepareSSEEvaluation() const
+    requires (has::prepareSSEEvaluation<SrcType> && Expression::CommonElement::isRealFloat)
+    {
+        src.prepareSSEEvaluation();
+
+        if constexpr (clampLow)
+            lowerBoundSIMD.sse = Expression::SSE::broadcast (lowerBound);
+
+        if constexpr (clampHigh)
+            upperBoundSIMD.sse = Expression::SSE::broadcast (upperBound);
+    }
+
+    VCTR_FORCEDINLINE VCTR_TARGET ("sse4.1") SSERegister<value_type> getSSE (size_t i) const
+    requires (archX64 && has::getSSE<SrcType> && Expression::allElementTypesSame && Expression::CommonElement::isRealFloat)
+    {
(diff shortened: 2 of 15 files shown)


benchmark/benchmark.cpp
JanosGit, line 1: These benchmarks are not run by default anywhere at the moment, you need to build and execute them manually

include/vctr/Expressions/DSP/Decibels.h
```diff
@@ -45,9 +58,25 @@ template <size_t extent, class SrcType, class DecibelConstant>
...
+template <size_t extent, class SrcType, class DecibelConstant, class MinDb>
+using FastMagToDb = ClampByConstant<extent,
+                                    MultiplyVecByConstant<extent,
+                                                          FastLog2<extent, SrcType>,
+                                                          ::vctr::detail::Log10ToLog2Constant<DecibelConstant>>,
+                                    MinDb,
+                                    DisabledConstant>;
+
+template <size_t extent, class SrcType, class DecibelConstant>
+using FastDBToMag = FastExp2<extent,
```
p4th4x0r: For consistency i'd name this FastDbToMag (small b letter)

include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -166,6 +172,201 @@ class FastExp : ExpressionTemplateBase
...
+    static constexpr int mantissaBits = 52;
+    static constexpr double minExpo = -1022.0; // exponent of minimum binary64 normal
+    static constexpr double expoBias = 1023.0; // binary64 exponent bias
+    static constexpr double a = -0x1.6e75d58p+2;
+    static constexpr double b = 0x1.bba7414p+4;
+    static constexpr double c = 0x1.35eccbap+2;
+    static constexpr double d = 0x1.f5e53c2p-2;
+    static constexpr double e = 1LL << mantissaBits;
+};
+}
```
p4th4x0r, line 206: image

include/vctr/Expressions/DSP/FastExp.h
```diff
@@ -166,6 +172,201 @@ class FastExp : ExpressionTemplateBase
...
+
+    VCTR_COMMON_UNARY_EXPRESSION_MEMBERS (FastExp2, src)
+
+    VCTR_FORCEDINLINE constexpr value_type operator[] (size_t i) const
+    {
+        auto p = src[i];
+
+        p = std::max (p, Constants::minExpo);
+
+        auto w = std::floor (p);
```
p4th4x0r, line 232: can we make sure that std::floor can be constant evaluated? afaik this is just available from c++26

include/vctr/Expressions/DSP/FastLog.h
```diff
@@ -0,0 +1,214 @@
...
+    static constexpr float a              { 1.1920928955078125e-7f };
+    static constexpr float b              { 124.22551499f };
+    static constexpr float c              { 1.498030302f };
+    static constexpr float d              { 1.72587999f };
+    static constexpr float e              { 0.3520887068f };
+    static constexpr int32_t c_0x007fffff { 0x007fffff }; // Masks the mantissa bits
+    static constexpr int32_t c_0x3f000000 { 0x3f000000 };
+    // clang-format on
+};
+}
```
p4th4x0r, line 44: I'm missing the constants specialization for double

include/vctr/Expressions/DSP/FastLog.h
```diff
@@ -0,0 +1,214 @@
...
+        auto xBitsInterpretedAsInt = std::bit_cast<int32_t> (x);
+        auto mantissa = (xBitsInterpretedAsInt & Constants::c_0x007fffff);
+        auto mx = std::bit_cast<float> (mantissa | Constants::c_0x3f000000);
+
+        auto y = static_cast<float> (xBitsInterpretedAsInt) * Constants::a;
+
+        auto dv = Constants::d / (Constants::e + mx);
+        auto ml = Constants::c * mx;
+
+        return y- Constants::b - ml - dv;
```
p4th4x0r: forgot a whitespace after y :)

test/TestCases/Expressions/Decibels.cpp
```diff
@@ -62,3 +63,37 @@ TEMPLATE_PRODUCT_TEST_CASE ("Decibels", "[VCTR][Decibels]", (PlatformVectorOps,
...
+    // Transforming back to linear values should result in the original values
+    const vctr::Vector magFS     = filter << vctr::fastDbToMag<vctr::dBFS> << dBFS;
+    const vctr::Vector magFSU    = filter << vctr::fastDbToMag<vctr::dBFS> << dBFS.template subSpan<1>();
+    const vctr::Vector magPower  = filter << vctr::fastDbToMag<vctr::dBPower> << dBPower;
+    const vctr::Vector magPowerU = filter << vctr::fastDbToMag<vctr::dBPower> << dBPower.template subSpan<1>();
+
+    REQUIRE_THAT (magFS,     vctr::Equals (srcD).withMargin (0.005));
+    REQUIRE_THAT (magFSU,    vctr::Equals (srcUnaligned).withMargin (0.005));
+    REQUIRE_THAT (magPower,  vctr::Equals (srcD).withMargin (0.005));
+    REQUIRE_THAT (magPowerU, vctr::Equals (srcUnaligned).withMargin (0.005));
```
p4th4x0r, line 98: Hm, for such functions i don't know if 0.005 is small enough as tolerance.

benchmark/benchmark.cpp
```diff
@@ -187,4 +188,30 @@ TEST_CASE ("Multiplication")
             return notAccelerated();
         };
     }
+}
+
+TEST_CASE ("Precise vs fast decibel benchmarks",  "[VCTR][Decibels]")
+{
+    auto srcMag16 = UnitTestValues<float>::array<16, 0, 0, 2> (false, true);
+    auto srcDb16 = UnitTestValues<float>::array<16, 0, -100, 10> (false, true);
```
p4th4x0r, line 196: maybe add some more values and a larger range? i.e. from -200 to +100 ?

include/vctr/Expressions/BasicMath/Clamp.h
p4th4x0r, line 1: Could you add a clamp overload that takes a class with getStart() and getEnd() function? so:
`template <class Range>
auto clamp (const Range& range) { return clamp (range.getStart(), range.getEnd()); }`
?

include/vctr/Expressions/DSP/Decibels.h
JanosGit: - [x] small b

benchmark/benchmark.cpp
JanosGit: As this is just a performance benchmark, I don't really expect that value ranges to matter

include/vctr/Expressions/BasicMath/Clamp.h
JanosGit: Good proposal, I'll add that to the backlog!

include/vctr/Expressions/DSP/FastExp.h
JanosGit: Actually, I'm not even sure if these bit trickeries work in a constant evaluated context. And there is also no real point in using this function in a constant evaluated context, you can just use the precise versions there. So I guess the real solution is to remove the `constexpr` specifier from this function

- [x] remove constexpr

include/vctr/Expressions/DSP/FastLog.h
JanosGit: This is because my reference implementation was float only. It's quite likely that we can implement something similar for double but as I'm not as deep into all of this to adapt the implementation myself, I limited it to float only for now. This is also modeled by the constraint in line 55 below.

include/vctr/Expressions/DSP/FastLog.h
JanosGit: - [x] add whitespace

test/TestCases/Expressions/Decibels.cpp
JanosGit: Yeah, I also figured that I mixed epsilon and margin based checks here.

I'd prefer to do a proper evaluation of the actual precision of these implementations in the course of a separate task and then adjust the test cases accordingly, before starting with try & error.

I added that to the backlog

p4th4x0r (approved):

PR #153, opened 2025-11-20
Documentation: Try to fix deployment to GH pages

JanosGit:


pmglr:
Can you elaborate? The title says "try", did you succeed?

JanosGit:
> Can you elaborate? The title says "try", did you succeed?

No I did not succeed yet. At some point I noticed that the version of the documentation published seems to be pretty outdated. I'm not sure if it has been updated at any point after initial publishing. However, I could verify that the new documentation gets built and published as expected to the gh-pages branch which should automatically be used for the website. I ran out of time tracking down this issue, so I'll revisit it once I find some time for that

pmglr:
> > Can you elaborate? The title says "try", did you succeed?
> 
> No I did not succeed yet. At some point I noticed that the version of the documentation published seems to be pretty outdated. I'm not sure if it has been updated at any point after initial publishing. However, I could verify that the new documentation gets built and published as expected to the gh-pages branch which should automatically be used for the website. I ran out of time tracking down this issue, so I'll revisit it once I find some time for that

Alright, thx for the explanation!

PR #157, opened 2025-12-12
vctr.h: Fix wrong license declaration in JUCE module section of the header

JanosGit:
Just stumbled across this inconsistency. Another goal of this PR is to check if the documentation is uploaded again to GH pages, since I changed some settings that hopefully make GitHub picking up the next changes pushed to the docu branch again.

### DIFF
--- a/include/vctr/vctr.h
+++ b/include/vctr/vctr.h
@@ -32,7 +32,7 @@
   name:               VCTR
   description:        A powerful C++ 20 wrapper around your favourite standard library containers.
   website:            https://github.com/sonible/vctr
-  license:            MIT
+  license:            LGPLv3
   minimumCppStandard: 20
   dependencies:
   OSXFrameworks:      Accelerate


iromur (approved):
Good, important fix!
Cannot see though whether the GH pages doc was actually updated or not, don't know where to look for the generation/upload time/reference commit? Maybe it would be good to include some version information with the documentation (e.g., dynamically on generation).

JanosGit:
> Cannot see though whether the GH pages doc was actually updated or not, don't know where to look for the generation/upload time/reference commit? Maybe it would be good to include some version information with the documentation (e.g., dynamically on generation).

I expect the new documentation to be published after merging the PR. Background: The configuration was set to some different setting than using the gh-pages branch for the repo page, which is the way that we want our docu build to work. The actual docu always gets build and pushed to that branch, it was just GitHub ignoring the branch until now because of the wrong setting. I did not see anything new appearing right after the change but I guess that's because a new site will only be published on changes on the branch.

Some quick research didn't reveal a simple command to make the generation time of the docs appear in a prominent place of the generated website, but I'm pretty sure that there is a way, let's have a look into that soon

JanosGit:
Update: Yes the settings change did the trick, the docs have been updated

PR #160, opened 2026-01-12
Expressions: add ElementwiseMin/Max

pmglr:


include/vctr/Expressions/BasicMath/ElementwiseMax.h
```diff
@@ -0,0 +1,86 @@
+/*
+  ==============================================================================
+    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+
+    Copyright 2022- by sonible GmbH.
```
JanosGit: That should probably be 2026 now?

test/TestCases/Expressions/ElementWiseMax.cpp
```diff
@@ -0,0 +1,36 @@
...
+                            "[VCTR][elementWiseMax]",
+                            (PlatformVectorOps, VCTR_NATIVE_SIMD),
+                            (float, double, int32_t, int64_t, uint32_t, uint64_t) )
+{
+    VCTR_TEST_DEFINES (10)
+
+    const vctr::Vector max = filter << vctr::elementwiseMax (srcA, srcB);
+    auto reference = vctr::Vector (srcA.size(), [&] (auto n) { return std::max (srcA[n], srcB[n]); });
+
+    REQUIRE_THAT (max, vctr::Equals (reference).withMargin (1e-6));
```
JanosGit: I would say that since no real computation is done here a check for exact euqality is the correct thing in this case

pmglr: You have a point there!

include/vctr/Expressions/BasicMath/ElementwiseMax.h
pmglr: I actually copied this from another file without reading it. This exact line is used countless times; does the "-" have any particular meaning or is it a typo?

JanosGit: I did a quick research on that topic.

It seems to be half-common legal consensus that the year in a source file license header should always reflect the year of writing the file, although there are projects that update the years in their file headers on every first January to the current year and other big companies don't mention a year at all. In general, the GPLv3 template license notice text just starts at line 7 and does not include this line, so we have added that line on our own.

I could also find more recent files in the project that mention 2023, 2024 and 2025 here, so it seems that we followed the pattern of always putting in the creation year. Given that the initial version has been released in 2022, still most of the files will have 2022 here.

Given that we decided to add years but never updated the headers on year changes in the past I come to the conclusion that based on my research putting 2026 here is probably the right thing.

Regarding the "-", I think that's a typo indeed that has sticked around in the project for far too long. Funnily, some of the more recent files still contain it while others have it removed. We should probably clear this mess up ;)

JanosGit (approved):

pmglr:
We have a few spurious test failures, most recently the fast dB approximations. I will check for issues...

include/vctr/Expressions/BasicMath/ElementwiseMax.h
pmglr: Thanks for looking into this. I already updated it to 2026 (without the "-")!

JanosGit:
> We have a few spurious test failures, most recently the fast dB approximations. I will check for issues...

Please add them to our internal test failure tracking sheet, there are already a few spurious VCTR test fails noted there, we should look at them in a structurized way

PR #164, opened 2026-01-13
Spurious test fails: increase margins

pmglr:
Review-ready now. I tested it with the following run config, using seeds that have provoked test failures in the past:

```
      - name: Run test (seeded for Add)
        run: ./build/test/TestMain/vctr_test --rng-seed 3043489734

      - name: Run test (seeded for fastExp)
        run: ./build/test/TestMain/vctr_test --rng-seed 1285096097

      - name: Run test (seeded for Decibels)
        run: ./build/test/TestMain/vctr_test --rng-seed 760312143

      - name: Run test (seeded for Mean)
        run: ./build/test/TestMain/vctr_test --rng-seed 1432506547
```

### DIFF
--- a/test/TestCases/Expressions/Map.cpp
+++ b/test/TestCases/Expressions/Map.cpp
@@ -40,7 +40,7 @@ T map (T val)
     return val;
 }
 
-TEMPLATE_PRODUCT_TEST_CASE ("Map", "[VCTR][expressions]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double))
+TEMPLATE_PRODUCT_TEST_CASE ("Map", "[VCTR][Expressions][map]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double))
 {
     constexpr int srcValueMin = -52;
     constexpr int srcValueMax = 4000;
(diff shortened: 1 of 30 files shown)


test/TestCases/Expressions/Map.cpp
```diff
@@ -40,7 +40,7 @@ T map (T val)
     return val;
 }
 
-TEMPLATE_PRODUCT_TEST_CASE ("Map", "[VCTR][expressions]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double))
+TEMPLATE_PRODUCT_TEST_CASE ("Map", "[VCTR][expressions][map]", (PlatformVectorOps, VCTR_NATIVE_SIMD), (float, double))
```
JanosGit: Oh I think this expressions tag should start with a capital letter to fully align with the tags that you added. Maybe do a quick search for other occurences of [expressions]?

JanosGit (commented):
Thank you so much for the test tag cleanup and the error margin adjustments. Really appreciate that. Too bad that another test now failed 🫣

pmglr:
> Thank you so much for the test tag cleanup and the error margin adjustments. Really appreciate that. Too bad that another test now failed 🫣

I think it was about time!

Be aware that this is a draft PR.

test/TestCases/Expressions/Map.cpp
pmglr: Maybe then I should also transform the `[Complex]` tags to upper case

pmglr:
It is strange that the test for Decibels doesn't fail anymore with the recorded seed... I didn't make any relevant changes in that file

.github/workflows/UnitTests.yml
```diff
@@ -142,5 +142,5 @@ jobs:
         run: chmod +x ./build/test/TestMain/vctr_test
         if: runner.os != 'Windows'
 
-      - name: Run test
-        run: ./build/test/TestMain/vctr_test
+      - name: Run test (Normalize Sum)
+        run: ./build/test/TestMain/vctr_test --rng-seed 2798420011
```
JanosGit: I assume you are going to revert this before merging?

JanosGit (changes requested):
Please revert the explicit random seed. Otherwise approved!

JanosGit (approved):

PR #166, opened 2026-02-23
SubSpan: Removed assertions when creating a subspan with extent 0

TheWaveWarden:


JanosGit:
So this means e.g. 
```
vctr::Vector v { 1, 2 };
auto s = v.subSpan (2);
```
Would now be valid and non-asserting in contrast to the previous version. It will simply return a span of size zero, right? I can see some sense in that ;)
Would call for some test cases but assertions are not really testable – at least at the moment. Did you still test some permutations locally to verify that all now asserts or does not assert as expected?

TheWaveWarden:
> So this means e.g.
> 
> ```
> vctr::Vector v { 1, 2 };
> auto s = v.subSpan (2);
> ```
> 
> Would now be valid and non-asserting in contrast to the previous version. It will simply return a span of size zero, right? I can see some sense in that ;) Would call for some test cases but assertions are not really testable – at least at the moment. Did you still test some permutations locally to verify that all now asserts or does not assert as expected?

Yes, I figured it would only be complete if we extend the zero spans to all creator options.

I just ran a series of tests:
```
        vctr::Array v { 1.0f, 2.0f };

        const auto s1 = v.subSpan (2);
        const auto s2 = v.subSpan<2>();

        const auto s3 = v.subSpan (2, 0);
        const auto s4 = v.subSpan<2, 0>();

        const auto s5 = v.subSpan (1, 0);
        const auto s6 = v.subSpan<1, 0>();

        const auto s7 = v.subSpan (1, 1);
        const auto s8 = v.subSpan<1, 1>();

        const auto s9 = v.subSpan (3);      // asserts
        // const auto s10 = v.subSpan<3>(); // doesn't compile

        const auto s11 = v.subSpan (2, 1);     // asserts
        // const auto s12 = v.subSpan<2, 1>(); // doesn't compile
```

JanosGit (approved):

PR #167, opened 2026-05-20
Schn 373

JanosGit:


include/vctr/Expressions/Exp/Exp.h
```diff
@@ -70,6 +70,16 @@ class Exp : ExpressionTemplateBase
...
     }
+
+    VCTR_FORWARD_PREPARE_SIMD_EVALUATION_UNARY_EXPRESSION_MEMBER_FUNCTIONS
+
+    VCTR_FORCEDINLINE NeonRegister<value_type> getNeon (size_t i) const
+    requires archARM && has::getNeon<SrcType> && std::same_as<value_type, float>
+    {
+        const auto in = src.getNeon (i);
+
+        return { vexpf (in.value) };
```
JanosGit: I just realized that I can pull of similar tricks on either intel macs or Windows if MSVC is used as it supports intel SVML intrinsics since Visual Studio 2019. Let me add that for completeness and an equally fast intel support

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+namespace vctr::expressions
+{
+template <size_t extent, class SrcType, is::binaryCaseDistinctionConfig CaseConfig>
+requires is::comparableBy<ValueType<SrcType>, CaseConfig::compareOp>
+class BinaryCaseDistinction : vctr::ExpressionTemplateBase
+{
+public:
+    using value_type = ValueType<SrcType>;
+
+
```
pmglr: extra blank line here

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+    static constexpr value_type threshold = CaseConfig::ThresholdValue::value;
+    mutable SIMDRegisterUnion<Expression> thresholdSIMDConst;
+
+public:
+    template <class Src>
+    constexpr BinaryCaseDistinction (Src&& s)
+        : src (std::forward<Src>(s)),
+          caseTrue (CaseConfig::caseTrue << src),
+          caseFalse (CaseConfig::caseFalse << src)
```
pmglr, line 81: Ok so this means that expressions are constructed for both cases and depending on the comparison with the threshold one of them will be evaluated?

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+template <class T>
+concept binaryCaseDistinctionConfig = requires (const T& t) { [] <const auto& b, CompareOp op, class Th, const auto& a> (const BinaryCaseDistinctionConfig<b, op, Th, a>&) {} (t); };
+
+}
+
+namespace vctr::expressions
+{
+template <size_t extent, class SrcType, is::binaryCaseDistinctionConfig CaseConfig>
+requires is::comparableBy<ValueType<SrcType>, CaseConfig::compareOp>
+class BinaryCaseDistinction : vctr::ExpressionTemplateBase
```
pmglr, line 58: Deserves a little documentation

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+        auto compareResult = Expression::Neon::template compare<CaseConfig::compareOp> (src.getNeon (i), thresholdSIMDConst.neon);
+
+        return Expression::Neon::bitwiseBlend (caseFalseResult, caseTrueResult, compareResult);
+    }
+};
+}
+
+namespace vctr
+{
+/** Combines two expressions. Which one is evaluated depends on the comparison of the input value with the threshold using the specified compare operation.
```
pmglr: This brief doc is a bit long. The second sentence reads more like a detailed explanation. How about 

```
Evaluates one of two expressions depending on a binary condition.

The comparison of the input value with the threshold using the specified compare operation
decides which one is evaluated.
```

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+
+    The expression will prepend the source to both caseTrueExpression and caseFalseExpression internally. If a source element compares true to threshold
+    using compareOp, the result element will be the evaluation result of caseTrueExpression, otherwise caseFalseExpression. If you only want to apply an
+    expression to either the true or false path, consider passing vctr::forward as expression to the other path.
+
+    Note: Since this expression evaluates both expression paths and the compare operation this leads to possibly multiple accesses of the source. Therefore,
```
pmglr: Btw there's also a `@note` command

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
```diff
@@ -0,0 +1,192 @@
...
+};
+}
+
+namespace vctr
+{
+/** Combines two expressions. Which one is evaluated depends on the comparison of the input value with the threshold using the specified compare operation.
+
+    The expression will prepend the source to both caseTrueExpression and caseFalseExpression internally. If a source element compares true to threshold
+    using compareOp, the result element will be the evaluation result of caseTrueExpression, otherwise caseFalseExpression. If you only want to apply an
+    expression to either the true or false path, consider passing vctr::forward as expression to the other path.
```
pmglr: I see now why you implemented an expression that does nothing!

pmglr: I wonder if it wouldn't be more handy to have an `evaluateIf` expression instead of passing `forward` for one of the two cases?

test/TestCases/Expressions/BinaryCaseDistinction.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+TEMPLATE_PRODUCT_TEST_CASE ("BinaryCaseDistinctions", "[VCTR][Expressions][abs]", (DontUseAcceleration, VCTR_NATIVE_SIMD), (float, double, int32_t, int64_t) )
+{
+    VCTR_TEST_DEFINES (10)
+
+    static constexpr auto threshold = ElementType (1);
+
+    // square values below 1, cube them above
+    const vctr::Vector cases = filter << vctr::binaryCaseDistinction<vctr::square, vctr::CompareOp::lessOrEqual, threshold, vctr::cube> << srcA;
```
pmglr, line 44: The order of the template parameters reminds me more of Python's ternary operator than C++'s

```
square(x) if x <= 1 else cube(x)
```

vs.

```
x<= 1 ? square(x) : cube(x)
```

Any particular reason you chose this order?

test/TestCases/Expressions/BinaryCaseDistinction.cpp
```diff
@@ -0,0 +1,47 @@
...
+template <isAnyOf<float, double, int32_t, int64_t> T>
+T squareBelowOneCubeAbove (T x)
+{
+    if (x < T (1))
+        return x * x;
+
+    return x * x * x;
+}
+
+TEMPLATE_PRODUCT_TEST_CASE ("BinaryCaseDistinctions", "[VCTR][Expressions][abs]", (DontUseAcceleration, VCTR_NATIVE_SIMD), (float, double, int32_t, int64_t) )
```
pmglr: test tag is "abs"

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
JanosGit: I went with the approach here that is used across the entire range of VCTR expression implementations that is adding the actual documentation to the ExpressionChainBuilder instance of it declared below. Rationale: The documentation that would be put here would be basically a duplicate of that one, however the instance below is what is user-code facing and therefore needs the detailed bit of information.

I'm open for discussion if this is really the best documentation style, but in case we decide to change the style I'd prefer to apply this change to all expression implementations in one general refactoring task.

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
JanosGit: Yes and no. Yes it is constructed for both cases, which should usually boil down to almost nothing in most cases.

For the element wise `operator[]` based evaluation, it will indeed work like that, comparing the source value to the threshold and depending on that result evaluation the one or the other for that element. 

For SIMD based evaluation on the other hand, it will be evaluated for both and both will be combined using a `bitwiseBlend` operation that takes a comparison result bitmask as third argument, leading to branch free code.

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
JanosGit: I like that, will change it!

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
JanosGit: Mhm, you might have a point here. I also like the expressiveness of an `evaluateIf` expression. I'd still prefer to put that onto the backlog for now and go with the solution implemented here, in order to not stretch the timebox for this task too much.

test/TestCases/Expressions/BinaryCaseDistinction.cpp
JanosGit: classic copy & paste artifact I'd say 😅

test/TestCases/Expressions/BinaryCaseDistinction.cpp
JanosGit: I shuffled around the order during development and that order felt most intuitive to me in the end. And to be honest, I feel the same for that python ternary operator which I prefer from a logical flow perspective over the C++ convention. But no more specific reason than that to be honest. Would you prefer a different order?

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
pmglr: Ok makes sense. @copydoc would be a nice option but it's not resolved by my IDE and therefore wouldn't be very helpful

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
pmglr: Sure. Do you want to create a task for it?

test/TestCases/Expressions/BinaryCaseDistinction.cpp
pmglr: I would prefer the C++ style order indeed but since there's no inherent "natural" order  but I'll leave the choice to the author.

pmglr (approved):

include/vctr/Expressions/Generic/BinaryCaseDistinction.h
JanosGit: I added one to the internal backlog

PR #168, opened 2026-08-17
Array: test different initialization syntaxes

pmglr:


JanosGit:
Your changes look good and sensible and should proof that initialisation works as intended with all syntax variations. But it seems like this flagged some rather CI related Win build issue. Therefore I'd propose to fix that first and leave this PR open for so long, ok?

pmglr:
> Your changes look good and sensible and should proof that initialisation works as intended with all syntax variations. But it seems like this flagged some rather CI related Win build issue. Therefore I'd propose to fix that first and leave this PR open for so long, ok?

of course!

JanosGit (approved):

PR #170, opened 2026-09-01
Vctr 176 GitHub actions build failure for all win builds

JanosGit:


JanosGit:
This was a tougher one than expected. Conan would always fail to detect the Visual Studio version running on the GitHub Windows runners. This is likely the side effect of a recent change in their runner image configuration.

I ended up with creating custom conan profiles for Windows builds with significant help from claude which also created some crazy vs version detection stuff in the actions workflow

.github/workflows/UnitTests.yml
```diff
@@ -10,91 +10,116 @@ jobs:
...
+        # Every config selects the Conan profile it wants to build with via `conan_profile`:
+        #   - `default`: let Conan auto-detect the profile
+        #   - anything else: use `.github/conan/profiles/<conan_profile>`
+        # See .github/conan/profiles/Readme.md for details.
         configs:
           - runner: ubuntu-22.04
             cc: clang-15
             cxx: clang++-15
             ipp: no_ipp
+            conan_profile: default
```
TheWaveWarden, line 22: So "default" is just the config that was used before?

.github/workflows/UnitTests.yml
```diff
@@ -123,18 +148,77 @@ jobs:
...
+            {
+              $installationPath = & $vswhere -latest -products * -property installationPath
+              $defaultToolsFile = Join-Path $installationPath "VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt"
+              if (Test-Path $defaultToolsFile) { $vcToolsVersion = (Get-Content $defaultToolsFile -Raw).Trim() }
+            }
+          }
+
+          # MSVC tools version 14.3x -> runtime v143 / msvc 193, 14.4x -> v144 / 194, 14.5x -> v145 / 195
+          $vsRuntimeVersion = "v143"
+          $msvcVersion = "193"
```
TheWaveWarden, line 176: I don't fully get what's going on here, but I see a lot of hardcoded version numbers. Assuming this has to be updated when runners change again?

TheWaveWarden (commented):
Looking good as far as I can see

.github/workflows/UnitTests.yml
JanosGit: Yes, this will make conan use it's self-generated default profile instead of an explicitly specified one. It's basically a conan term used here

.github/workflows/UnitTests.yml
JanosGit: As already stated, this was written by claude, such powershell stuff is definetively not my main area of expertise 😁 

However what we see here is the initialization to some sane default values, if you follow the code below you see that it tries to find out the actual versions and will print a message that it's falling back to the defaults in case it doesn't find any known version. What needs to be updated here in future is probably the known values in line 180.

If this wasn't a github hosted runner, I'd be happy with just pinning the values in the conan profile to known values, but I think this is actually a clever solution in the context of constantly changing build environments round here

TheWaveWarden (approved):

